PR #44, opened 2020-07-15
Convert generation script to node.js

anthmatic:
Ports the generation logic from python to node. Can be run using `npm run generate`.

This also adds some special handling for block tags to lookup their end tag from their snippets and use that. This works, but I think it would be better to address this in the future by updating the actual jinjavadoc annotations to include a properly formatted snippet.


williamspiro:
I know this is WIP but I believe you will need to add this file to https://github.com/HubSpot/HubL-Language-Extension/blob/javascript-port/.vscodeignore so the extension knows this file is not part of the package of files used for the extension functionality itself.

anthmatic:
I'm looking to get this in shortly to get some other work out.

 Feel free to comment here if you anyone has anything they'd like to see changed or addressed. I also created the https://github.com/HubSpot/HubL-Language-Extension/tree/python-v0.1 branch to hang on to the original Python version if needed

PR #57, opened 2020-07-30
Update extension info for launch

### LINKED ISSUE #46: Add contributing guide
... and instructions for developing/debugging the extension locally

### LINKED ISSUE #58: Integrate syntax highlighting into readme and highlight top level features
Syntax highlighting for HubL in html, css and JS should be called out somewhere in the very begging of the README.

Maybe something like:
>This is a HubL language extension for the Visual Studio Code IDE, allowing for 🚀 fast local HubSpot CMS Platform development in .html files. This extension supports:
>- HubL syntax highlighting for HTML, CSS, and JavaScript files
>- Autom-complete snippets for HubL tags, filters, functions, operators, expression tests and more
>- indentation...
>- HubL dleimeter auto-complete... - HubL comments - ...etc.
>For comprehensive HubL documentation, see the HubL docs.

We can also remove 
>For nice HubL syntax highlighting, install the Jinja scopes extension which lays on top of .html files.

anthmatic:
After merging, we'll need to:
- update this repo name (to `hubspot-cms-vscode`)
- build/publish the [extension to the marketplace](https://code.visualstudio.com/api/working-with-extensions/publishing-extension)

Resolves #58 
Resolves #46

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright 2019  HubSpot, Inc.
+Copyright 2020  HubSpot, Inc.
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,12 @@
-MIT License
+Copyright 2019  HubSpot, Inc.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
 
-Copyright (c) 2018 William Spiro
+   http://www.apache.org/licenses/LICENSE-2.0
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
-
-The above copyright notice and this permission notice shall be included in all
-copies or substantial portions of the Software.
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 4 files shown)


LICENSE
```diff
@@ -1,21 +1,12 @@
...
-copies or substantial portions of the Software.
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-SOFTWARE.
+Copyright 2019  HubSpot, Inc.
```
anthmatic, line 1: @williamspiro any concerns with changing the type of license for this?

gcorne: Update to 2020?

package.json
```diff
@@ -14,9 +14,9 @@
     },
     "repository": {
         "type": "git",
-        "url": "https://github.com/williamspiro/HubL-Language-Extension"
+        "url": "https://github.com/HubSpot/HubL-Language-Extension"
     },
-    "homepage": "https://github.com/williamspiro/HubL-Language-Extension/blob/master/README.md",
+    "homepage": "https://github.com/HubSpot/HubL-Language-Extension/blob/master/README.md",
```
gcorne, line 19: I wonder if we should rename the repo. Maybe `hubspot-vscode-extension`, `hubspot-cms-vscode-extension`, or `hubspot-cms-vscode`?

anthmatic: I think we should wait to get ownership of the marketplace listing, though

LICENSE
williamspiro: I think it is fine, but thanks for checking 👍

anthmatic:
@gcorne @williamspiro This should be all set to get merged and publish the extension. I can do this later on today.

I'm not sure where @TheWebTech is regarding documenting this and not sure if we want to coordinate publishing with having docs available?

TheWebTech:
@anthmatic It's already in the developer workflow doc. I will be adding some more call outs to it elsewhere, like in the CLI reference page there will be a mention of it.

I don't recall where, but I had previously asked if we thought it made sense to effectively transfer the documentation of this extension from the extension readme to our developer docs. My thinking was that this is more similar to our API clients, where it is platform dependant(the platform here being VS Code) versus the CLI, which isn't. Based on that it may make sense to keep the core "how to use" information in the repository, which is also used for the marketplace listing, where people may find it without ever having seen it on our developer docs.

I have a changelog announcement for this written up - that needs to go out after this is merged and live in the VS Code marketplace. I just need to know when this goes live.

anthmatic:
Thanks, @TheWebTech. I think the way it is documented works for now. We can definitely revisit that in the future once we have a better idea of how folks are using it, etc...

williamspiro:
This is awesome and exciting! 🚢  sounds good to me (and then stop working 😄)

PR #61, opened 2020-08-07
[WIP/POC] Add ability to customize syntax highligting via settings

anthmatic:
If a user has installed a VS Code theme that affects syntax highlighting color schemes, it is possible that the HubL highlighting might not work correctly (it _does work_ but colors might look the same for different types).

This PR aims to do two things:
- changes the file type from `HTML` to `HTML+HUBL` (still associated with `.html` files)
- adds settings to for users to override syntax highlighting colors used
image

This is really just a proof of concept  at this stage but I'm wondering what folks might think about these changes?
@TanyaScales @TheWebTech @williamspiro

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,12 @@
-MIT License
+Copyright 2019  HubSpot, Inc.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
 
-Copyright (c) 2018 William Spiro
+   http://www.apache.org/licenses/LICENSE-2.0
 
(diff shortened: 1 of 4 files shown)


TheWebTech:
Interesting solution to the problem. One part I don't fully understand, will it treat all HTML files as HTML+HubL?

Seems like that could break other extensions that queue off of HTML.

That said if it only happens as the result of a user directly enabling it or changing the file type in VS Code to HTML+HubL I think that'd be okay.

TanyaScales:
> changing the file type in VS Code to HTML+HubL

I can see a user mapping their `.html` files to the HTML+HubL language type to enable the syntax highlighting for these files. That was the way the BetterJinja extension was working for syntax highlighting.

PR #67, opened 2020-09-01
Removed HubL language configuration overrides for HTML and CSS filetypes

miketalley:
- Language configuration is now only applied to `HubL` filetypes instead of `HTML` and `CSS`
- Improved display of `HubL` language type from `hubl` to `HubL (hubl)` in VSCode selection dropdown

HubL filetype comment
image

HTML filetype comment
image

CSS filetype comment
image

HubL selection in filetype dropdown
image

/cc @williamspiro


miketalley:
> How is HubL vs normal HTML file detected? Is it just manually when the user selects it from the language dropdown?

Either that or if the file contains `.hubl.html` extension([configuration link](https://github.com/HubSpot/hubspot-cms-vscode/blob/8cb8bc604b891c330bc17caf35c0bdd617c2886a/package.json#L83)).

I think some of the issues that we were seeing were caused by the fact that we were piggy-backing the HubL language configuration onto all HTML files. This PR removes that piggy-backing so that the configuration for HubL files only applies to HubL files. I left the syntax highlighting for HubL within HTML files([configuration link](https://github.com/HubSpot/hubspot-cms-vscode/blob/issue/65-hubl-comment-overrides/package.json#L95-L99)) alone because I don't believe it will break anything.

There will be a bit more effort needed to get the functionality for HubL files because they will need to be manually identified in VSCode. Some ways that we can simplify this in the future would be...
- Have a configuration file for projects(`.hsconfig`?) that we can use to trigger adding this functionality to `.html` files within the project using [workspaceContains](https://code.visualstudio.com/api/references/activation-events#workspaceContains)
- Use a separate extension for HubL filetypes(like `.hubl.html`)

I couldn't find another way to do this currently, if anyone has any other ideas that would be great.

williamspiro:
The only thing I am realizing playing with this branch locally is the HubL snippets only seem to work in HubL files.

anthmatic:
Thanks for working on this @miketalley. 

In the past, I wasn't a big fan of the `hubl.html` and `hubl.css` convention but these days I do think like that might be a good direction to move in.

Alternatively, I think we could use the `firstline` property: https://code.visualstudio.com/api/references/contribution-points#language-example to support some sort of pragma based approach.

miketalley:
> The only thing I am realizing playing with this branch locally is the HubL snippets only seem to work in HubL files.

Good point Will, I just pushed a commit to allow snippets in both HTML and HubL filetypes.

package.json
```diff
@@ -23,10 +23,54 @@
...
+            {
+                "language": "hubl",
+                "path": "./snippets/man_gen/hubl_email_variables.json"
+            },
+            {
+                "language": "hubl",
+                "path": "./snippets/man_gen/hubl_required_email_template_variables.json"
+            },
             {
                 "language": "html",
```
anthmatic, line 75: should these get moved to "hubl" too?

anthmatic (approved):

PR #73, opened 2020-09-15
Support html-hubl and css-hubl

anthmatic:
Prior to this PR, we the extension was active and affecting all `.html` and `.css` files, regardless of whether they were intended to be HubL files. This adds the following languages:

- `hubl (raw)`: most likely won't be used, but required for the following types
- `html-hubl`: HTML + HubL
- `css-hubl`: CSS + HubL

To clarify some confusion around the way this works: This implements these new languages as "injection grammars" in VS Code, which is the recommended way to _extend_ and existing grammar (like CSS or HTML). An alternative is to create an embedded language, but that's meant more for cases where you want to fence off an area of code (think `<style>` blocks in HTML).


anthmatic:
FYI, I'm writing up documentation for this, but I wanted to give folks a chance to check this out now

syntaxes/css-hubl.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "css-hubl",
+  "scopeName": "text.css.hubl",
+  "comment": "HubL CSS Templates",
+  "patterns": [
+    {
+      "include": "source.css"
+    }
```
gcorne, line 8: Is there a reason we're not using`{ "include": "source.hubl" }` here as a second pattern like we are in `html-hubl` and are instead using the `injectionSelector`?

package.json
```diff
@@ -23,96 +23,97 @@
     "icon": "images/hubspot-logo.png",
     "main": "./main.js",
     "activationEvents": [
-        "onLanguage:hubl"
+        "onLanguage:html-hubl",
+        "onLanguage:css-hubl"
```
gcorne, line 27: Should we move the indention rules into the language config?

syntaxes/css-hubl.json
anthmatic: Yeah, the HTML syntax plays a lot nicer with the HubL syntax (which I believe is because of the way tags get tokenized). CSS, on the other hand, doesn't and so HubL needs to be injected.

Injection should work for HTML, I'll test that out real quick to see if we can clean that up.

package.json
anthmatic: Good question. Previously that did not work but I'll try that again now that we're not hijacking existing language types.

syntaxes/css-hubl.json
gcorne: Can you elaborate on what doesn't work?

package.json
anthmatic: Thanks for pointing that out, that worked. 

I also removed `main.js` the `activationEvents` for now. The docs suggest that you do need `activationEvents` but the extension does not need to work properly (tested, works for hubl types and not original langs). Also, their example language configuration project does not use them: https://github.com/microsoft/vscode-extension-samples/blob/master/language-configuration-sample/package.json

package.json
```diff
@@ -124,34 +120,45 @@
...
             {
                 "language": "hubl",
                 "scopeName": "source.hubl",
-                "path": "./syntaxes/hubl.json"
+                "path": "./syntaxes/hubl.json",
+                "injectTo": [
+                    "text.html.hubl",
+                    "text.css.hubl"
+                ]
             },
```
miketalley, line 145: Do we need this grammar anymore? Can we just use `html-hubl` and `css-hubl` to prevent confusion when selecting the file type?

anthmatic: As of right now the `hubl` grammar gets injected into `html-hubl` and `css-hubl` so it is still needed. But based on @gcorne's comment above, I want to dig into this to make sure that it's set up correctly.

PR #74, opened 2020-09-16
Update readme with recent changes

anthmatic:
I was originally planning to make some smaller updates for #73 but I wanted to split this out because I wanted to see about making the read me more concise.

Also adds changelog entry

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ content.absolute_url
 }
 ```
 
-Other helpful snippet include:
+Other helpful snippets include:
 | Snippet Prefix | Description | Example |
 |-|-|-|
 | otrue | Generates overrideable=True for HubL tags | overrideable=True |
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ _Thanks to the [Better Jinja](https://github.com/samuelcolvin/jinjahtml-vscode) 
 - Block comment toggling: Press `CMD + /` to create HubL comments
 - Block level indentation: Text inside of `{% block %}` will be indented automatically
 
-### ubL Snippets
+### HubL Snippets
 All HubL supported tags, filters, expression tests and functions have auto-complete snippets. Expression tests are accessed by typing the test name alone, filters are accessed with `|` and functions/tags are accessed with `~`. All snippets include descriptions and parameter details. You up/down arrow to navigate the IntelliSense and hit enter to execute a snippet. Snippet completed HubL statements will auto-highlight available parameters, which can be tabbed through (`${parameter}`). Additionally there is support for other commonly needed syntax, such as module fields.
 
 ## Getting Started
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ If you're new to CMS Hub, check out how to [get started with local development.]
 ## Features
 
 ### Syntax Highlighting
-HubL syntax hightlighting is supported for `.HTML` and `.CSS` files. To enable syntax highlighting, make sure that your file type is set to `HTML + HUBL` or `CSS + HUBL`, respectively.
+HubL syntax highlighting is supported for `.html` and `.css` files support HubL syntax highlighting. Enable syntax highlighting by setting your file type to HTML + HubL or CSS + HubL.
 
 _Thanks to the [Better Jinja](https://github.com/samuelcolvin/jinjahtml-vscode) extension for inspiration._
 
--- a/README.md
+++ b/README.md
@@ -40,6 +40,19 @@ hubdb_table_rows(${table_id}, ${query})
 content.absolute_url
 ```
 
+There are also other helpful snippets:
+| Snippet Prefix | Description | Example |
+|-|-|-|
+| otrue | Generates overrideable=True for HubL tags | overrideable=True |
+| ofalse | Generates overrideable=False for HubL tags | overrideable=False |
+| for | Returns a basic if statement | {% for {iterable} in {dict} %} {{ iterable }} {% endfor %}" |
+| if | Returns a basic if statement | {% if {test} %} do_something}{% endif %} |
+| elif | Else if statement to be used within if statement | {% elif {test} %} |
+| else | Else statement to be used within if statement | {% else %} |
+| hubldoc | Boilerplate html/HubL document |  |
+| hublblog | Boilerplate blog markup |  |
+
+
 ## Getting Started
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -20,143 +20,16 @@ _Thanks to the [Better Jinja](https://github.com/samuelcolvin/jinjahtml-vscode)
...
-```
-group.group
-```
-To create a repeater, use
-```
-group.repeater
-```
-all associated fields need to be added into the children array within the snippet output
+## Getting Started
+This extension introduces new file languages: `HTML + HUBL` and `HTML + CSS`. In order to use these new features, you'll want to make sure you're told VSCode to use these langauges. For one-off files, you can simple change the language in the lower right hand corner of your screen. However, we recommend adjusting your file associations under your User or Workspace preferences:
```
miketalley, line 27: should be: you'll make sure `you've`
and: you can `simply`

README.md
```diff
@@ -20,143 +20,16 @@ _Thanks to the [Better Jinja](https://github.com/samuelcolvin/jinjahtml-vscode)
...
+- In VSCode, press `CMD` + `SHIFT` + `P` to open the command prompt
+- Search for `files.associations`
```
miketalley, line 29: This does not work for me. I just get a "No matching commands" when searching for this
image

Maybe we can direct them to open the user/project settings and add it to the JSON? Also a link to https://code.visualstudio.com/docs/getstarted/settings might be useful here.

anthmatic: Ah, good catch! LMK what you think about the updated verbiage. I reference the UI because you can easily tab between User/Workspace vs the JSON method which involves two different files in two different locations.

README.md
```diff
@@ -1,5 +1,5 @@
 # HubL Visual Studio Code Language Extension
```
TheWebTech, line 1: Let's officially rename it in the readme HubSpot Visual Studio Code Extension or HubSpot VS Code Extension.
While it primarily is used to make HubL easier to work with, it provides other utilities like JSON snippets for fields.
I could see the feature list growing to encompass more HubSpotty things.

README.md
```diff
@@ -11,7 +11,7 @@ If you're new to CMS Hub, check out how to [get started with local development.]
 ## Features
 
 ### Syntax Highlighting
-HubL syntax hightlighting is supported for `.HTML`, `.CSS`, and `.JS` files. To enable syntax highlighting, make sure that your file type is set to one of these types.
```
TheWebTech, line 14: Highlighting is spelled wrong.

Also suggesting this alternate:

> `.html` and `.css` files support [HubL](https://developers.hubspot.com/docs/cms/hubl) syntax highlighting. Enable syntax highlighting by [setting your file type](https://code.visualstudio.com/docs/languages/overview#_changing-the-language-for-the-selected-file) to `HTML + HubL` or `CSS + HubL`.

README.md
```diff
@@ -20,143 +20,20 @@ _Thanks to the [Better Jinja](https://github.com/samuelcolvin/jinjahtml-vscode)
...
 - Block level indentation: Text inside of `{% block %}` will be indented automatically
 
-### __HubL Snippets__
-All HubL supported tags, filters, expression tests and functions have auto-complete snippets. Expression tests are accessed by typing the test name alone, filters are accessed with `|` and functions/tags are accessed with `~`. All snippets include descriptions and parameter details. You up/down arrow to navigate the IntelliSense and hit enter to execute a snippet. Snippet completed HubL statements will auto-highlight available parameters, which can be tabbed through (`${parameter}`).
+### HubL Snippets
+All HubL supported tags, filters, expression tests and functions have auto-complete snippets. Expression tests are accessed by typing the test name alone, filters are accessed with `|` and functions/tags are accessed with `~`. All snippets include descriptions and parameter details. You up/down arrow to navigate the IntelliSense and hit enter to execute a snippet. Snippet completed HubL statements will auto-highlight available parameters, which can be tabbed through (`${parameter}`). Additionally there is support for other commonly needed syntax, such as module fields.
```
TheWebTech, line 24: This section becomes a little hard to read. Maybe after first sentence bullet how to access each auto completion?

> A prefix is required to access some auto completions:
> 
> - Expression tests are accessed by typing the test name alone.
> - filters are accessed with `|`.
> - functions/tags are accessed with `~`.

Also noticed typo. "You"

>All snippets include descriptions and parameter details. Use `up`/`down` arrow keys to navigate the [IntelliSense](https://code.visualstudio.com/docs/editor/intellisense) and hit `enter` to execute a snippet.

README.md
```diff
@@ -20,143 +20,20 @@ _Thanks to the [Better Jinja](https://github.com/samuelcolvin/jinjahtml-vscode)
...
-```
-[__HubL Functions__](https://designers.hubspot.com/en/docs/hubl/hubl-supported-functions) produce entire HubL function statements with available parameters, without wrapping curly braces. The intention of this is so you can use HubL functions within other HubL statements easily (like setting variables, for loops, etc.) Ex `~hub` > Enter produces:
-```
-hubdb_table_rows(${table_id}, ${query})
-```
-[__HubL Expression Tests__](https://designers.hubspot.com/docs/hubl/operators-and-expression-tests#expression-tests) produce expression test names. Ex `di` > Enter produces:
-```
-divisibleby
-```
+## Getting Started
```
williamspiro, line 26: This does not seem to be quite how it works for me. I need to search for something more like "Files: Associations" to get the setting to show up. I also see "Edit in settings.json" rather than "add item"

Screen Shot 2020-09-18 at 9 42 48 AM

README.md
```diff
@@ -20,143 +20,20 @@ _Thanks to the [Better Jinja](https://github.com/samuelcolvin/jinjahtml-vscode)
...
-[`Email Required Template Variables`](https://designers.hubspot.com/docs/hubl/hubl-supported-variables#required-email-template-variables)
-```
-{{ site_settings.company_city }}
-{{ site_settings.company_name }}
-{{ unsubscribe_link }}
-{{ unsubscribe_link_all }}
-etc.
-```
-
-__Custom Module Fields__
```
TheWebTech, line 144: I think adding custom module fields back is to the readme is kind of a must, otherwise it's not mentioned at all.

README.md
anthmatic: Which version of VS Code are you running?

williamspiro: I was on 1.30.1 , after upgrading this is working now

anthmatic: I figured that was the case. Looks like the UI was updated in July 2019 so I'm not too worried about documenting the old way unless anyone disagrees

PR #77, opened 2020-09-16
Validate JSON schema

anthmatic:
Exploring the idea of adding field JSON schemas to this extension. 

Adding these mean that we could validate specific files to make sure that required properties and types are correct: image

### DIFF
--- a/schema/module-fields.json
+++ b/schema/module-fields.json
@@ -13,10 +13,15 @@
     {
       "type": "object",
       "required": ["type"],
-      "properties": { "type": { "enum": ["boolean", "choice"] } },
+      "properties": { "type": { "enum": ["boolean", "choice", "color", "cta", "datetime", "email", "date"] } },
       "allOf": [
          {"if":{"properties":{"type":{"const":"boolean"}}}, "then":{"$ref": "./boolean.json"}},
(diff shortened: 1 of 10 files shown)


schema/module-fields.json
```diff
@@ -0,0 +1,28 @@
...
+      "required": ["type"],
+      "properties": { "type": { "enum": ["boolean", "choice", "color", "cta", "datetime", "email", "date"] } },
+      "allOf": [
+         {"if":{"properties":{"type":{"const":"boolean"}}}, "then":{"$ref": "./boolean.json"}},
+         {"if":{"properties":{"type":{"const":"choice"}}}, "then":{"$ref": "./choice.json"}},
+         {"if":{"properties":{"type":{"const":"color"}}}, "then":{"$ref": "./color.json"}},
+         {"if":{"properties":{"type":{"const":"cta"}}}, "then":{"$ref": "./date.json"}},
+         {"if":{"properties":{"type":{"const":"datetime"}}}, "then":{"$ref": "./datetime.json"}},
+         {"if":{"properties":{"type":{"const":"email"}}}, "then":{"$ref": "./email.json"}},
+         {"if":{"properties":{"type":{"const":"date"}}}, "then":{"$ref": "./date.json"}}
```
anthmatic, line 24: Relatively simple to keep fields separate and compose them so we could also have a similar setup for theme fields.

PR #83, opened 2020-10-15
Add tracking using `@hubspot/cms-lib`

gcorne:


anthmatic (approved):
Looks good to me. The only thing I was thinking is we might want to add an setting to turn usage tracking off, but we I don't think we need that since requires a config

gcorne:
Yeah, I was thinking that the existing config option for disabling tracking was enough.

PR #95, opened 2020-12-02
Add inline error linting

anthmatic:
Adds inline template errors:
image

Todo:

- [x] Debounce file change event
- [x] Add some transparency around auth / config
- [x] Refactor this into it's own module

Also need to dig deeper into issues with validation results:
- [x] Missing or negative `startPosition` values
- [x] Differences between warning levels
- [x] How to handle errors for imports/extends


index.js
```diff
@@ -33,8 +35,51 @@ async function activate() {
...
+      new vscode.Position(error.lineno - 1, error.startPosition - 1)
+    );
+  } else {
+    return document.lineAt(new vscode.Position(error.lineno - 1, 0)).range;
+  }
+};
+
+async function updateDiagnostics(document, collection) {
+  console.log('Request');
+  const { renderingErrors } = await validateHubl(getPortalId(), document.getText());
```
anthmatic, line 66: would be nice to debounce or throttle this

lib/lint.js
```diff
@@ -0,0 +1,70 @@
...
+  const templateErrors = fitleredErrors.map((error) => {
+    return {
+      code: '',
+      message: error.message,
+      range: getRange(document, error),
+      severity: vscode.DiagnosticSeverity.Error,
+    };
+  });
+
+  if (document) {
```
brandenrodgers, line 34: I'm pretty new to this vs code repo, but I think you can probably bail early in this function if document doesn't exist. It would save you from having to unnecessarily do the validateHubl, filter, and map above.

index.js
```diff
@@ -28,11 +30,38 @@ async function activate() {
...
+        'ACTIVATION',
+        {},
+        getAccountId()
+      );
+    } catch (e) {
+      console.log(e);
+    }
+  }
+
+  if (vscode.workspace.getConfiguration('hubl').get('lint.enabled')) {
```
brandenrodgers, line 50: Just a suggestion (feel free to ignore), but you could move `hubl` and also `lint.enabled` to constants since they're referenced in a few places.

lib/lint.js
```diff
@@ -0,0 +1,70 @@
...
+    updateDiagnostics(vscode.window.activeTextEditor.document, collection);
+  }
+
+  documentChangeListener = vscode.workspace.onDidChangeTextDocument(
+    (editor) => {
+      if (editor) {
+        clearTimeout(timeout);
+        timeout = setTimeout(function () {
+          updateDiagnostics(editor.document, collection);
+        }, 1000);
```
brandenrodgers, line 57: I played around with it and the 1 second debounce feels pretty good to me, but I don't really have a good idea of what the standard is for linting. Have you looked into it? We should make sure we don't accidentally put out something that will cause vs code to work too hard.

lib/lint.js
```diff
@@ -0,0 +1,70 @@
...
+};
+
+async function updateDiagnostics(document, collection) {
+  const { renderingErrors } = await validateHubl(
+    getPortalId(),
+    document.getText()
+  );
+
+  const fitleredErrors = renderingErrors.filter((error) => {
+    return error.reason != 'MISSING';
```
brandenrodgers, line 22: Do we have consts defined for the error reasons? Or is that defined in the BE?

lib/lint.js
anthmatic: I took a sample of some other extension that lint code and it's all over the place (250-1000ms) but most plugins don't denounce at all because the linter process is running on the local machine instead of needing to make an API request.

lib/lint.js
anthmatic: AH, good catch. I'm trying to remember why I even added this, I feel like I might have intended to check something else. I'll take a second look.

lib/lint.js
anthmatic: Defined on BE, but since you pointed it out I added the possible values to `constants.js`. I'm mainly concerned with focusing out `missing` files for now and although VS Code provides different severity levels, the HS `templateError` types all seem like syntax `Error`s

anthmatic:
Still working the kinks out, but based on some user feedback I've thought I'd attempt to track down missing files instead of just supressing those errors. I also started to distinguish between different error severities:
image

anthmatic:
Made some updates to address some user feedback. The biggest change is that I now attempt to check whether a file is present or not, so paths used in tags like `import` are now validated

lib/constants.js
```diff
@@ -0,0 +1,27 @@
...
+  DISABLED: 'Information',
+  SYNTAX_ERROR: 'Error',
+  UNKNOWN: 'Error',
+  BAD_URL: 'Warning',
+  EXCEPTION: 'Error',
+  MISSING: 'Warning',
+  OTHER: 'Information',
+};
+
+const WORD_REGEX = /{%.*(.*).*%}/;
```
brandenrodgers, line 21: What does this do exactly? Might be a good idea to add a comment to explain it so people who are bad at regex (like me 😅 ) can quickly understand.

anthmatic: Good call, I will update this name and add a comment

lib/validation.js
```diff
@@ -0,0 +1,86 @@
...
+    };
+  });
+
+  collection.set(document.uri, templateErrors);
+};
+
+let timeout = null;
+const triggerValidate = (document, collection) => {
+  if (!['html-hubl', 'css-hubl'].includes(document.languageId)) {
+    return clearValidation(document, collection);
```
brandenrodgers, line 77: Do we also want to call `clearTimeout(timeout)` here? I'm not sure if it's possible, but I'm imagining a scenario where [this collection.set](https://github.com/HubSpot/hubspot-cms-vscode/pull/95/files#diff-419296896634fcf982f1a147262973ede2ee05a726db022c357eafa6c6ed7ecfR71) gets called after we call `clearValidation` here b/c of the 1s timeout OR possibly b/c the `validateHubl` call could take long.

I haven't played around with this too much yet, so this race condition might not actually exist.

lib/validation.js
```diff
@@ -0,0 +1,86 @@
...
+const clearValidation = (document, collection) => {
+  collection.set(document.uri, null);
+};
+
+const updateValidation = async (document, collection) => {
+  if (!document) {
+    return collection.clear();
+  }
+
+  const { renderingErrors } = await validateHubl(
```
brandenrodgers, line 47: What happens when the `validateHubl` request fails or takes a long time?

package.json
```diff
@@ -165,10 +165,20 @@
                 "scopeName": "text.css.hubl",
                 "path": "./syntaxes/css-hubl.json"
             }
-        ]
+        ],
+        "configuration": {
+            "title": "HubL",
```
brandenrodgers, line 170: This "title" is the main title that shows at the top of settings, right? If so, should we rename it to match the name of our extension? I looked around a little bit and that seemed to be a fairly common pattern.

index.js
```diff
@@ -28,11 +30,38 @@ async function activate() {
...
 
-  if (!validateConfig() || !isTrackingAllowed()) {
+  if (!validateConfig()) {
     return;
   }
 
-  await trackUsage('vscode-extension', 'ACTIVATION');
+  if (isTrackingAllowed()) {
+    try {
+      await trackUsage(
```
brandenrodgers, line 39: Out of the scope of this PR, but it seems like we could/should update `trackUsage` to only require the first two args here. The rest seem somewhat unnecessary.

lib/validation.js
```diff
@@ -0,0 +1,86 @@
...
+    getPortalId(),
+    document.getText()
+  );
+
+  const fitleredErrors = renderingErrors.filter((error) => {
+    if (
```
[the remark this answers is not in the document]
anthmatic: I updated this name, essentially this tries to resolve file paths and filters out any files that DO exist. Open to feedback to make this clearer!

package.json
anthmatic: Good point. Since the name of the extension is: "HubSpot VS Code Extension", I updated this to be "HubSpot", to be consistent with other extensions:
image

PR #108, opened 2021-06-05
Show notification to user if Beta hasn't been enabled

anthmatic:
To spread awareness of Beta Features, this adds a notification that shows up for users to alert them of the new feature. Adds a button so that they can opt in without visiting settings.

image

Note: VSCode doesn't seem to allow any styling (or even new lines) in messages.

### DIFF
--- a/lib/notify.js
+++ b/lib/notify.js
@@ -1,6 +1,6 @@
 const { window } = require('vscode');
 
-const HAS_SEEN_BETA = 'has-seen-beta';
+const HAS_SEEN_BETA = 'HS_HAS_SEEN_BETA';
 const BETA_NOTIFICATION_MESSAGE =
   'Now in Beta: HubL Error Linting. Opt-in to beta feautures to enable this feature.';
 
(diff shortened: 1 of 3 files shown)


brandenrodgers:
Can you include a screenshot of the new notification?

lib/notify.js
```diff
@@ -0,0 +1,25 @@
+const { window, workspace, ConfigurationTarget } = require('vscode');
+const { EXTENSION_CONFIG_NAME, EXTENSION_CONFIG_KEYS } = require('./constants');
+
+const HAS_SEEN_BETA = 'HS_HAS_SEEN_BETA';
```
brandenrodgers, line 4: Maybe `HAS _SEEN_BETA` can live in the constants file? Maybe in some `GLOBA_STATE_KEYS` object.

brandenrodgers (approved):
Looks great! (pending a screenshot)

Also, one thing to consider is how we will show this message again to users when we release new beta features. Maybe it's worth updating the global key to be `HS_HAS_SEEN_BETA-1` or something like that

anthmatic:
Thanks @brandenrodgers. I updated the state key to `HS_HAS_SEEN_LINTING_MESSAGE`. I'm not 100% sure what future notif logic will look like but this should give us room to add more.

lib/notify.js
anthmatic: Good idea!

Issue #109, opened 2021-06-13
Extension causes high cpu load


GonzaloTorreras:
- Issue Type: `Performance`
- Extension Name: `hubl`
- Extension Version: `0.10.0`
- OS Version: `Windows_NT x64 10.0.19041`
- VS Code version: `1.56.2`

Got a +3 to 5 secs loading this extension on VSC.
:warning: Attached report file for inspection if needed:  [HubSpot.hubl-unresponsive.cpuprofile.txt](https://github.com/HubSpot/hubspot-cms-vscode/files/6644061/HubSpot.hubl-unresponsive.cpuprofile.txt)

Find more details here: https://github.com/microsoft/vscode/wiki/Explain-extension-causes-high-cpu-load

anthmatic:
Thank you, @GonzaloTorreras. Looking into this.

anthmatic:
@GonzaloTorreras I made some improvements in 0.13.0 if you're able to update to that, hopefully that solves this! If you're still having issues, feel free to open a new issue up!

PR #112, opened 2021-06-22
Bundle extension

anthmatic:
In #109, a user reported slow startup time. This bundles the extension files and node_modules, improving startup time. Also adds the boilerplate VSCode TS config to bring it more in line with standard extensions.


anthmatic:
Before:
image

After: 
image

PR #115, opened 2021-07-23
add style field snippets

anthmatic:
Adds snippets for new style fields. These snippets do not include examples of the default property since these fields are meant to typically be used as overrides, rather than the source of truth for a style.

### DIFF
--- a/snippets/man_gen/hubl_custom_module_fields.json
+++ b/snippets/man_gen/hubl_custom_module_fields.json
@@ -491,8 +491,8 @@
     "prefix": "styleTab.group",
     "body": [
       "{",
-      "\"name\": \"${1:style}\",",
-      "\"label\": \"${2:Style}\",",
+      "\"name\": \"${1:styles}\",",
+      "\"label\": \"${2:Styles}\",",
       "\"children\": [${3}],",
       "\"type\": \"group\",",
       "\"tab\": \"STYLE\"",
--- a/snippets/man_gen/hubl_custom_module_fields.json
+++ b/snippets/man_gen/hubl_custom_module_fields.json
@@ -486,5 +486,111 @@
       "}"
     ],
     "description": "HubSpot Font Field"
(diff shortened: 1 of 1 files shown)


snippets/man_gen/hubl_custom_module_fields.json
```diff
@@ -486,5 +486,111 @@
       "}"
     ],
     "description": "HubSpot Font Field"
+  },
+  "Style Tab Group": {
+    "prefix": "styleTab.group",
+    "body": [
+      "{",
+      "\"name\": \"${1:style}\",",
+      "\"label\": \"${2:Style}\",",
```
jasonnrosa, line 495: Nit thing but would it make sense to swap label and name to be Styles so that it matches the label on the current tab?

snippets/man_gen/hubl_custom_module_fields.json
```diff
@@ -486,5 +486,111 @@
...
+  },
+  "Alignment Field": {
+    "prefix": "field.alignment",
+    "body": [
+      "{",
+      "\"name\": \"${1:alignment_field}\",",
+      "\"label\": \"${2:Alignment field}\",",
+      "\"required\": false,",
+      "\"locked\": false,",
+      "\"type\": \"boolean\",",
```
jasonnrosa, line 511: Would it make sense to use the alignment field vs. a custom boolean field for this one (e.g. https://developers.hubspot.com/en/docs/cms/building-blocks/module-theme-fields#alignment)?

jasonnrosa:
Just added a couple of comments but besides those everything was looking good to me!

snippets/man_gen/hubl_custom_module_fields.json
anthmatic: Good catch!

anthmatic:
Thanks @jasonnrosa!

PR #117, opened 2021-08-06
Pass template context to validate

anthmatic:
Different validations apply to different template types. This PR allows parses the filepath and annotations (if they exist) to pass extra context along to the validate endpoint.

Note: this relies on future additions to cli-lib

### DIFF
--- a/package.json
+++ b/package.json
@@ -186,8 +186,7 @@
         }
     },
     "dependencies": {
-        "@hubspot/cli": "^3.0.7-beta.5",
-        "@hubspot/cli-lib": "^3.0.7-beta.5",
+        "@hubspot/cli": "^3.0.7",
         "fs-extra": "^9.0.1",
         "node-fetch": "^2.6.0"
     },
--- a/package.json
+++ b/package.json
@@ -186,7 +186,8 @@
         }
     },
     "dependencies": {
-        "@hubspot/cli-lib": "3.0.4",
+        "@hubspot/cli": "^3.0.7-beta.5",
+        "@hubspot/cli-lib": "^3.0.7-beta.5",
         "fs-extra": "^9.0.1",
         "node-fetch": "^2.6.0"
     },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
     "name": "hubl",
     "displayName": "HubSpot VS Code Extension",
     "description": "HubSpot CMS Hub support for VS Code",
-    "version": "0.11.0",
+    "version": "0.13.0",
```
anthmatic, line 5: hm.

package.json
```diff
@@ -186,7 +186,8 @@
         }
     },
     "dependencies": {
-        "@hubspot/cli-lib": "3.0.4",
+        "@hubspot/cli": "^3.0.7-beta.5",
+        "@hubspot/cli-lib": "^3.0.7-beta.5",
```
anthmatic, line 190: When https://github.com/HubSpot/hubspot-cli/pull/534/files is merged, will pin to newest release

src/lib/validation.js
```diff
@@ -39,21 +48,50 @@ const clearValidation = (document, collection) => {
...
 
+const getTemplateType = (document) => {
+  if (isCodedFile(document.fileName)) {
+    const annotations = getAnnotationsFromSource(document.getText());
+    return {
+      is_available_for_new_content:
+        getAnnotationValue(
+          annotations,
+          ANNOTATION_KEYS.isAvailableForNewContent
+        ) != 'false',
```
miketalley, line 72: Code looks solid and makes sense, just wondering if we can improve the way that `getAnnotationValue` works to make usage a bit cleaner 🤔

brandenrodgers: Yeah, I never really loved my implementation of that one. I just had one idea though, what if instead of `getAnnotationsFromSource`, we had something like `buildAnnotationValueGetter`. Basically the output would be a `getAnnotationValue` function that doesn't require the annotations to be passed into it. 
For example:

```js
const getAnnotationsFromSource = source => {
const annotations = //blah blah blah
return (...args) => getAnnotationValue(annotations, ...args);
}
```

anthmatic: Thats a cool idea. Changed that up here https://github.com/HubSpot/hubspot-cli/pull/534/commits/42459f358c58d5c37108d007036bfdce4aa989eb

PR #118, opened 2021-08-18
Move hubl linting out of beta

anthmatic:
Removes the existing `hubspot.beta` setting and introduces `hubspot.hublLinting`, which is turned on by default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -177,6 +177,11 @@
         "configuration": {
             "title": "HubSpot",
             "properties": {
+                "hubspot.beta": {
+                    "type": "boolean",
+                    "default": false,
+                    "markdownDescription": "Enable HubSpot VS Code extension beta features\n\n*Note: beta features may require a valid `hubspot.config.yml` file*"
+                },
                 "hubspot.hublLinting": {
                     "type": "boolean",
                     "default": true,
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -177,10 +177,10 @@
         "configuration": {
             "title": "HubSpot",
             "properties": {
-                "hubspot.beta": {
+                "hubspot.hublLinting": {
```
anthmatic, line 180: We _could_ keep this, but it would do nothing

package.json
```diff
@@ -177,10 +177,10 @@
...
                     "type": "boolean",
-                    "default": false,
-                    "markdownDescription": "Enable HubSpot VS Code extension beta features\n\n*Note: beta features may require a valid `hubspot.config.yml` file*"
+                    "default": true,
```
anthmatic, line 182: If someone had previously opted out of the beta, this will re-enable this feature for them. It's possible to check if they've opted out but based on a previous convo, I think we decided to just go ahead and enable for all.

package.json
miketalley, line 180: 🏆

miketalley (approved):
Looks great, I think having more specific beta features is awesome!

package.json
brandenrodgers: I like the idea of keeping it so that the pattern is easy to follow for the next beta feature that we'd like to introduce

brandenrodgers:
LGTM!

Issue #123, opened 2021-10-31
Add browser extension functionality if possible


TheWebTech:
Adding web extension support would enable the extension to work in github.dev as well as vscode.dev, as well as potentially other VS Code based code editors like gitpod and possibly codeanywhere. 

There certainly may be some functionality that wouldn't be supported in the web extension version. But a lot of it likely will work.

https://code.visualstudio.com/api/extension-guides/web-extensions#update-existing-extensions-to-web-extensions

brandenrodgers:
While this is a cool idea, it's not very realistic that we'll have the bandwidth to put into this to get it to work. We rely on some core libraries that depend on having access to the user's filesystem (like the cli config file). These libraries are not web compatible. Closing this for now, but maybe we can revisit in the future.

PR #124, opened 2021-11-16
WIP: Get extension working in vscode.dev

anthmatic:
Adds a new webpack config targeting browsers and fallbacks to handle most deps

Known issues:
- Config lookup: This logic lives in `@hubspot/cli-lib` and uses node's `fs` module so we'll need to find an alternative method of loading a config

Fixes: #123


src/lib/validation.js
```diff
@@ -40,7 +39,7 @@ const isFileInWorkspace = (error) => {
     filePath = filePath + '.module';
   }
 
-  return fs.existsSync(path.resolve(dirToActiveFile, filePath));
+  return vscode.workspace.fs.stat(path.resolve(dirToActiveFile, filePath));
```
anthmatic, line 42: Roughly how to address `fs`, however this extension relies on `cli-lib` which uses `fs` directly

anthmatic:
Just posting an update here on the progress:

Most of the functionality is here to allow the extension to work in the web version of VS Code and removes its reliance on the HS CLI-LIB as well. However, because the extension operates in a WebWorker, requests that are made to the HS API fail because they do not have CORS enabled. Discussing a path forward internally.

PR #134, opened 2022-10-28
[WIP] Extension exploration testing

miketalley:
- Adds custom HubSpot `viewsContainers` to the `activitybar`
- Adds `hubspot:portals` `viewsWelcome` with placeholder text and button to open personal-access-keys-ui
- Registers and displays tree data with names of portals listed in `hubspot.config.yml` if present
- Registers and displays tree data with links to documentation


anthmatic:
This is looking pretty cool!

src/lib/servers/auth.ts
```diff
@@ -0,0 +1,85 @@
+import * as vscode from 'vscode';
```
anthmatic, line 1: this is really cool! I think it would be worth thinking about bringing this to cli-lib (or a seperate pkg) so that CLI could benefit from this flow as well!

miketalley: Yeah, good call! If we end up using this, it would make a lot of sense to have it live in cli-lib.

.gitignore
```diff
@@ -114,3 +114,7 @@ dist
 # End of https://www.toptal.com/developers/gitignore/api/node
 
 .DS_store
+
+hubspot.config.yml
+/tmp
+/temp
```
bmatto, line 120: What was creating a `temp` directory?

miketalley: Me 😂  Do you think we should remove this?

bmatto: Yeah, likely not a common pattern ;).

src/extension.ts
```diff
@@ -0,0 +1,62 @@
...
+      );
+      vscode.commands.executeCommand(COMMANDS.PORTALS_REFRESH);
+    }
+  );
+
+  startServer({
+    onPostRequest: async (req: any) => {
+      return handleHubspotConfigPostRequest(req, { rootPath });
+    },
+  });
```
anthmatic, line 61: Is this always running? I think it would be best to only start this server up when needed

package.json
```diff
@@ -25,11 +26,9 @@
         "Snippets"
     ],
     "icon": "images/hubspot-logo.png",
-    "main": "./dist/index.js",
+    "main": "./dist/extension.js",
```
anthmatic, line 29: 👍

src/lib/auth.ts
```diff
@@ -0,0 +1,170 @@
...
+    env = 'prod',
+    name,
+    personalAccessKeyResp: { encodedOAuthRefreshToken: personalAccessKey },
+  } = req.body;
+  let configPath = loadHubspotConfigFile(rootPath);
+
+  if (configPath) {
+    setConfigPath(configPath);
+  } else {
+    configPath = `${rootPath}/hubspot.config.yml`;
```
anthmatic, line 136: I think we need need make sure Window style paths are supported here too

TanyaScales:
One thing I came across during testing -- in the event that I already have a PAK set up within a portal, but I don't have a config file set up anywhere. When going to authenticate, we should run a check for an already existing PAK on the `personal-access-key` page -- and bring that back to the extension. Otherwise, the developer will get stuck here in the flow since, it appears, that nothing happens in this scenario. (Please correct me if I'm wrong!)

package.json
```diff
@@ -188,23 +187,98 @@
...
+                    "title": "HubSpot",
+                    "icon": "images/hubspot-logo.svg"
+                }
+            ]
+        },
+        "views": {
+            "hubspot-explorer": [
+                {
+                    "id": "hubspot.treedata.portals",
+                    "name": "Portals",
```
TanyaScales, line 238:
```suggestion
                    "name": "Accounts",
```

I think we'll want to call this pane the `Accounts` pane. This is the language used throughout the docs and the HS app to refer to portals generally. So we should stay aligned with that.

package.json
TanyaScales, line 29:
```suggestion
    "main": "./dist/extension.ts",
```

miketalley:
> One thing I came across during testing -- in the event that I already have a PAK set up within a portal, but I don't have a config file set up anywhere. When going to authenticate, we should run a check for an already existing PAK on the `personal-access-key` page -- and bring that back to the extension. Otherwise, the developer will get stuck here in the flow since, it appears, that nothing happens in this scenario. (Please correct me if I'm wrong!)

Thanks for pointing this out! I had initially thought about this, but it got lost along the way. I will update this!

src/lib/auth.ts
```diff
@@ -0,0 +1,110 @@
...
+  console.log(
+    'loadedHubspotConfig',
+    config.defaultPortal,
+    JSON.stringify(config.portals.map((p: Portal) => p.name))
+  );
+
+  if (!validateConfig()) {
+    throw new Error(`Invalid config could not be loaded: ${path}`);
+  } else {
+    onLoadHubspotConfig();
```
anthmatic, line 58: Could we just call `updateConfigDependentFeatures()` directly here?

src/lib/auth.ts
```diff
@@ -0,0 +1,110 @@
...
+  rootPath,
+  onConfigFound,
+  onConfigUpdated,
+}: {
+  rootPath: string;
+  onConfigFound: Function;
+  onConfigUpdated: Function;
+}) => {
+  loadConfigDependentFeatures = onConfigFound;
+  updateConfigDependentFeatures = onConfigUpdated;
```
anthmatic, line 100: I might not be following this correctly but are we hoisting these functions to later be called by `initializeHubspotConfigDependents()`? If so, could they get defined in the same scope and called directly?

src/lib/auth.ts
miketalley: Yes, I had some different code in `onLoadHubSpotConfig` before that was removed, but left it for later use, but it can be removed now.

src/lib/auth.ts
miketalley: This is clearly more confusing than it needs to be. I refactored the callbacks into `vscode.command`s. I think it should be an improvement.

src/lib/statusBar.ts
```diff
@@ -0,0 +1,40 @@
...
+export const updateStatusBarItems = () => {
+  console.log('updateStatusBarItems');
+
+  const config = getConfig();
+  const defaultAccount = config.defaultPortal;
+
+  if (defaultAccount) {
+    hsStatusBar.text = `$(arrow-swap) ${defaultAccount}`;
+    hsStatusBar.tooltip = `Active HubSpot Account: ${defaultAccount}`;
+    hsStatusBar.backgroundColor = undefined;
```
anthmatic, line 17: I'm getting a TS error for that this property doesn't exist. We should probably move up the version that we support but I'm unsure what that should be

miketalley: Good call, I'm going to update the `@types/vscode` version.

anthmatic:
@miketalley I bumped into an issue testing out the auth flow where the vscode:// link opened up in the wrong VS Code window (which was already open). In this case it did nothing because the extension wasn't active but in production the extension would be active and have created a config in the wrong project. Is there any way we can open the window that initiated the auth flow?

miketalley:
> @miketalley I bumped into an issue testing out the auth flow where the vscode:// link opened up in the wrong VS Code window (which was already open). In this case it did nothing because the extension wasn't active but in production the extension would be active and have created a config in the wrong project. Is there any way we can open the window that initiated the auth flow?

I'll take a look into this in the AM! Thanks for letting me know!

PR #135, opened 2022-11-10
Adding Module Creation

miketalley:
# Description
Allows a module to be created within the extension UI

## Access Points
Right clicking in the file menu now has "Create new module" menu option
image

A folder/subfolder can also be right clicked to nest the module
image

## Flow/Demo
https://www.loom.com/share/21432166d01043ac85ea07342260e0e6

## Validation
- Adds `.module` to folder names that are provided without it.
- Checks if the module folder already exists when entering a folder name. If it exists, incrementally tries filenames in the format `<folderName><incrementedNumber>.module`.
  - For example if I try to create a folder `Foo` and there is already a `Foo.module` folder, the created folder name will be `Foo1.module`.

### Notes
Utilizes the `explorer.newFolder` command ([code ref link](https://github.com/microsoft/vscode/blob/6aaf830b9cfacd7e100fc0e5a59bf924580db75d/src/vs/workbench/contrib/files/browser/fileActions.ts#L919-L924)). The handler code can be found [here](https://github.com/microsoft/vscode/blob/6aaf830b9cfacd7e100fc0e5a59bf924580db75d/src/vs/workbench/contrib/files/browser/fileActions.ts#L835).

/cc @anthmatic @TanyaScales @jsines


bmatto:
During the flow - does the are the module question prompts one after another ... i.e. the user answers, hits enter, answers another, hits enter? I ask because I wonder if it would be better/possible to have all of those in a single form?

src/lib/commands/create/module.ts
```diff
@@ -0,0 +1,131 @@
...
+        contentTypes: string[] | undefined;
+        global: boolean | undefined;
+      } = {
+        moduleLabel: '',
+        contentTypes: [],
+        global: false,
+      };
+      let moduleName: string | undefined;
+
+      await showNamePrompt(destinationPath).then(async (name) => {
```
bmatto, line 107: With async/await do we still need to nest in callbacks like this?

bmatto:
Interesting, I read through the various UX APIs that VSCode is exposing. My inclination is that the picker is tough because the user can't go back in the flow. A form with all the fields allows the user to jump around and see all their values before they submit. While I appreciate the "don't use webview for a wizard" ... I'm not sure this is a wizard though, more of a webform. It's not really a setting either.

TanyaScales:
I have conflicting opinions about either direction -- I'm wondering if we can do something lower touch than this even and cut the form initially altogether to get an even more expedited module creation. In that we'd effectively cut the form and populate the module directory with a blank `label` and `host_template_types`. And once the `create new module` menu item is selected, we drop their cursor to name the directory (like native file creation) at that time.

The developer can do into the file structure at that time to adjust those values. We could even do a placeholder value for the label initially for all new modules, just to have something there.

This leaves a gap in parity from the CLI to the extension -- but I also think there is probably more to be done on this front thinking about the way we want users to give us information in the UI since this is only one of many creation commands that we are going to be working on. There are bound to be various lengths of questions/forms, different data types, etc. We should try to think about all those instances ahead of time to make sure we have all our use cases in sight before choosing a path here.

anthmatic:
+1 for narrowing the scope here and skipping the wizard. 
>And once the create new module menu item is selected, we drop their cursor to name the directory (like native file creation) at that time.

this would be slick and native feeling if we could do it

> We could even do a placeholder value for the label initially for all new modules, just to have something there.

We could probably leverage snippets for these. So each new file would get prepopulated with a snippet that has placeholders for all of the common stuff. Snippets have a bunch of autocomplete magic (types and enums and stuff). This extension does something similar: https://github.com/Gruntfuggly/auto-snippet

src/lib/fileHelpers.ts
```diff
@@ -0,0 +1,76 @@
...
+  );
+};
+
+/*
+ * Returns a non-duplicate folder name with the format <folderName>.module
+ * It will add a number to the end of the folder name if it already exists
+ * @param folderName - name of a folder with or without a .module extension
+ */
+const getUniqueModuleFolderName = (folderPath: string) => {
+  const folderName = folderPath.split('/').pop() || '';
```
anthmatic, line 24: Can we use `path.sep` here?

jsines (approved):
Not sure what's causing the folder to remain after the rename. Otherwise everything looks good!

Docs for `FileWillCreateEvent` say "To make modifications to the workspace before the files are created, call the [waitUntil](https://code.visualstudio.com/api/references/vscode-api#FileWillCreateEvent.waitUntil)-function with a thenable that resolves to a [workspace edit](https://code.visualstudio.com/api/references/vscode-api#WorkspaceEdit)."

Could it be that the `applyEdit` in 
```
vscode.workspace.applyEdit(edit).then(() => {
  copySampleModuleFilesToFolder(uniqueModulePath);
  resolve(edit);
});
```
is redundant and causing problems? It seems to work fine if the folder is named with `.module` so I'm sure the issue is somewhere within that try block

miketalley:
> Not sure what's causing the folder to remain after the rename. Otherwise everything looks good!
> 
> Docs for `FileWillCreateEvent` say "To make modifications to the workspace before the files are created, call the [waitUntil](https://code.visualstudio.com/api/references/vscode-api#FileWillCreateEvent.waitUntil)-function with a thenable that resolves to a [workspace edit](https://code.visualstudio.com/api/references/vscode-api#WorkspaceEdit)."
> 
> Could it be that the `applyEdit` in
> 
> ```
> vscode.workspace.applyEdit(edit).then(() => {
>   copySampleModuleFilesToFolder(uniqueModulePath);
>   resolve(edit);
> });
> ```
> 
> is redundant and causing problems? It seems to work fine if the folder is named with `.module` so I'm sure the issue is somewhere within that try block

If I disable the `applyEdit` call, the folder will simply get created, there is no rename done and the sample files do not get populated. I'm noticing that the `applyEdit` seems to return `false`, I'm looking into if this means that it failed somewhere. 👀

src/lib/fileHelpers.ts
```diff
@@ -0,0 +1,78 @@
...
+          ) {
+            cleanupCallback();
+            const uniqueModulePath = getUniqueModuleFolderName(
+              e.files[0].fsPath
+            );
+
+            edit.renameFile(e.files[0], vscode.Uri.file(uniqueModulePath));
+
+            vscode.workspace.applyEdit(edit).then(() => {
+              copySampleModuleFilesToFolder(uniqueModulePath);
```
anthmatic, line 66: Something tells me that the file is not renamed yet here and this function has an `ensureDirSync()` call which could be tripping up this edit.

Have you considered writing the module to the original path and changing the folder name after?

miketalley: It looks like it was a problem with `resolve(edit)` in two aspects..

1. When calling `resolve` on the `waitUntil` promise with a truthy value, it seems to break things with the edit
2. The call to `resolve` when no edit takes place should be a `reject`

PR #139, opened 2022-11-21
Update cli-lib and pick up fix for template annotation parse regression

anthmatic:
In the current build, the HubL linting feature falls over due to regression in `cli-lib` [that has been fixed](https://github.com/HubSpot/hubspot-cli/pull/770). 

This also bumps the version of cli-lib to ^4 while we're at it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -191,7 +191,8 @@
         }
     },
     "dependencies": {
-        "@hubspot/cli-lib": "^3.0.7",
+        "@hubspot/cli": "^4.1.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -191,7 +191,8 @@
         }
     },
     "dependencies": {
-        "@hubspot/cli-lib": "^3.0.7",
+        "@hubspot/cli": "^4.1.3",
```
anthmatic, line 194: todo: remove

PR #146, opened 2022-12-02
[WIP] Autocomplete for file paths

anthmatic:
Autocomplete for file paths!
image

Super-POC stage but it looks very doable so I'll be cleaning this up and pushing forward on it.

This will apply to these HubL tags/params:
- `include`
- `import`
- ` module path=""` - should also support built-in module paths
- ... tbd

### DIFF
--- a/src/lib/providers.ts
+++ b/src/lib/providers.ts
@@ -1,5 +1,6 @@
 import * as vscode from 'vscode';
-
+import * as fs from 'fs';
+import * as path from 'path';
 import { AccountsProvider } from './providers/accountsProvider';
 import { COMMANDS, TREE_DATA } from './constants';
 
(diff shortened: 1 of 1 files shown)


TanyaScales (approved):

lolcabanon:
Hey @anthmatic !

I don't really know where else to submit my idea, so i'll do it here, since it is a little bit on the same vibe. Feel free to point me in the right direction if there is a better place to discuss that!

I am currently building a CMS Theme starting from the _wonderful_ `cms-theme-boilerplate` and the experience is mostly awesome! *Congrats to HubSpot team for this btw!*

Here is an idea I got and would love to see implemented :

**Common paths shorthand/alias** a bit like the `@hubspot/module-name` functionality and inspired by the SvelteKit `$lib` alias ([see doc for details](https://learn.svelte.dev/tutorial/lib)).

The possibility to use custom aliases like `@theme/module-name` or `@modules/module-name` in templates would be amazing instead of the current `../../modules/module-name`. At least predefined aliases for common paths like "modules" or "sections" would be good enough and probably easier to manage on your end (if the project structure follows a standard pattern, of course).

I also understand that paths like `@hubspot/module-name` seems to be absolute paths to HubSpot built-in modules, so it might not be just "add another alias" but rather adding a completely new alias feature.

Anyway, just wanted to start the discussion and see where it would go! :smile:

jsines:
Hi @lolcabanon ! 👋 

We always love hearing feedback on these tools so your contribution is well appreciated. 

I've filed two separate issues based on your comment to get them on our dev team's radar. Feel free if you'd like to add additional info to those tickets, otherwise you can just watch them for any updates.

https://github.com/HubSpot/hubspot-cms-vscode/issues/252
https://github.com/HubSpot/hubspot-cms-vscode/issues/253

Thanks for the suggestion & bug report!

PR #147, opened 2022-12-02
[WIP] HubL Language Server

anthmatic:


anthmatic:
Right now, this isn't a viable path forward because VS Code is responsible for triggering autocomplete and it doesn't trigger inside of strings. 

Continuing on with #146

PR #150, opened 2022-12-06
[POC] View remote files

anthmatic:
I was curious to dig into some of the viewTree / file handling stuff and put together this POC that lists remote files from the design manager and allows you to view them read-only: 
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -253,6 +253,11 @@
                     "id": "hubspot.treedata.quickLinks",
                     "name": "Quick Links"
                 }
+                ,
+                {
+                    "id": "hubspot.treedata.remoteFs",
+                    "name": "Remote FS"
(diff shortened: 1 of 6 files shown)


miketalley:
This is awesome!

TanyaScales:
This is pretty cool -- I wonder if we could expand on this POC and make a right click context menu that would "Create local copy" -- (basically the fetch command) -- and plop that into a specified area.

It would be pretty cool to just include this as is (read only state) though for devs -- I imagine just having the reference close by and in the same format as they are working locally would help them diff files and work more efficiently in one space.

TanyaScales:
Closing out this pull request as the extension now supports viewing remote files via @jsines PRs

PR #152, opened 2022-12-07
Uri handler multiple windows

### LINKED ISSUE #151: Auth flow does not specify which project to open vscode:// link in
When using the auth flow the vscode:// link will open in the active VSCode window. There is a potential for the wrong window to be in context when the data for auth is returned, resulting in creating a config in the wrong project.

- Look into a way we can open the window that initiated the auth flow

/cc @anthmatic

miketalley:
Fixes #151 

Related Info
https://code.visualstudio.com/api/references/vscode-api#window.registerUriHandler

### DIFF
--- a/src/lib/commands/auth.ts
+++ b/src/lib/commands/auth.ts
@@ -11,7 +11,10 @@ import { loadHubspotConfigFile } from '../auth';
 let configFoundAndLoaded = false;
 let hubspotConfigWatcher: fs.FSWatcher | null;
 
-export const registerCommands = (context: vscode.ExtensionContext) => {
+export const registerCommands = (
+  context: vscode.ExtensionContext,
+  rootPath: string
+) => {
   context.subscriptions.push(
     vscode.commands.registerCommand(
       COMMANDS.ON_CONFIG_FOUND,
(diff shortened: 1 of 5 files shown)


anthmatic:
Testing this locally (with the PAKUI changes locally) it still just seems to open the most recent window, not the originating window. Am I missing a step?

src/lib/commands/auth.ts
```diff
@@ -1,11 +1,13 @@
...
-export const registerCommands = (context: ExtensionContext) => {
+export const registerCommands = (
+  context: ExtensionContext,
+  rootPath: string
+) => {
   context.subscriptions.push(
     commands.registerCommand(COMMANDS.AUTHORIZE_ACCOUNT, async () => {
-      const authUrl =
-        'https://app.hubspot.com/l/personal-access-key/auth/vscode';
+      const authUrl = `https://app.hubspot.com/l/personal-access-key/auth/vscode${rootPath}`;
```
anthmatic, line 10: I can't fully test the flow on Windows but since `rootPath` uses `fsPath`, this path contains the windows style path which looks like `.../auth/vscodec:\folder` which causes a 404

anthmatic (approved):
Can't fully test the Windows stuff yet but LGTM otherwise

src/lib/commands/auth.ts
miketalley: Switched to using `path` instead of `fsPath` and also added a fix for the value of `rootPath` that is returned from personal-access-key-ui, which is something like `/C:/Folder`. The fix slices off the leading `/`.

Let me know if you see any issues with this.

PR #154, opened 2022-12-08
Rename trackAction -> trackEvent, adding data, more tracked events

miketalley:
- Adds user setting for extension `allowUsageTracking`
- Updates `trackAction` to only work if all usage tracking settings are enabled
  - `telemetry.enableTelemetry`
  - `hubspot.allowUsageTracking`
  - `allowUsageTracking` within `hubspot.config.yml`
- Adds user identification information to tracked events
  - extensionName (`hubspot.hubl`)
  - source (`appName` which is `hubl` for the extension currently)
  - vscodeVersion (version of VSCode being used)
  - extensionVersion (version of VSCode extension being used)
  - language (user's set language within VSCode)
  - machineId (unique machine identifier)
  - os (user's operating system and version)
  - shell (shell being used by default for user)
- Adds tracked events
```
export const TRACKED_EVENTS = {
  ACTIVATE: 'activated',
  AUTH_INITIALIZE_CONFIG: 'authInitializedConfig',
  AUTH_UPDATE_CONFIG: 'authUpdatedConfig',
  AUTHORIZE_ACCOUNT_CLICKED: 'authorizeAccountClicked',
  CREATE_MODULE: 'createdModule',
  DELETE_ACCOUNT: 'accountDeleted',
  LINTING_ENABLED: 'lintingEnabled',
  LINTING_DISABLED: 'lintingDisabled',
  RENAME_ACCOUNT: 'accountRenamed',
  RENAME_ACCOUNT_ERROR: 'accountRenameError',
  SELECT_DEFAULT_ACCOUNT: 'selectedDefaultAccount',
  SET_DEFAULT_ACCOUNT: 'setDefaultAccountClicked',
  UPDATE_DEFAULT_ACCOUNT: 'defaultAccountUpdated',
};
```


package.json
```diff
@@ -187,6 +187,11 @@
                     "type": "boolean",
                     "default": true,
                     "markdownDescription": "Enables code linting for HTML+HUBL and CSS+HUBL languages.\n\n*Requires a valid `hubspot.config.yml` file*"
+                },
+                "hubspot.allowUsageTracking": {
+                    "type": "boolean",
+                    "default": true,
+                    "markdownDescription": "Allow usage tracking within extension."
```
anthmatic, line 194: Could we tie this option to the setting in the `hubspot.config.yml`? Or maybe just rely on config + VS Code setting? I think having 3 separate ways to opt in/out of tracking will get confusing.

src/extension.ts
```diff
@@ -1,16 +1,23 @@
...
+import { initializeTracking, trackEvent } from './lib/tracking';
+import { TRACKED_EVENTS } from './lib/constants';
 
 export const activate = async (context: ExtensionContext) => {
-  console.log('Activating Extension...');
+  await trackEvent(TRACKED_EVENTS.ACTIVATE);
+  console.log(
+    'Activating Extension Version: ',
+    // @ts-ignore TODO - Why is extension not available, when it is?
+    context.extension.packageJSON.version
```
anthmatic, line 19: It wasn't available until v1.55: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52117/files#diff-0a5f3d01be8ab95a0ca83baa6915417d7497d254e6f0b508447db7388111a2e4R6082

I think it's worth bumping the supported version to 1.55

src/lib/commands/config.ts
```diff
@@ -18,18 +19,24 @@ const showRenameAccountPrompt = (accountToRename: Portal) => {
     .showInputBox({
       placeHolder: 'Enter a new name for the account',
     })
-    .then((newName: string | undefined) => {
+    .then(async (newName: string | undefined) => {
```
anthmatic, line 22: nit: could this just be optional?

miketalley: I'm not sure that I follow. When we are renaming an account, we need the `newName`, right?

anthmatic: sorry I just mean could `newName: string | undefined` be `newName?: string`

package.json
jsines: Agreed, even having the two between `hubspot.config.yml` and VSCode is a little confusing, but unavoidable if we want to use cli-lib functionality. I think we can stick to the two and remove this

src/extension.ts
```diff
@@ -1,16 +1,23 @@
...
 import { initializeConfig } from './lib/auth';
 import { initializeTerminal } from './lib/terminal';
-import { registerEvents } from './lib/events';
+import { initializeTracking, trackEvent } from './lib/tracking';
+import { TRACKED_EVENTS } from './lib/constants';
 
 export const activate = async (context: ExtensionContext) => {
-  console.log('Activating Extension...');
+  initializeTracking(context);
+  await trackEvent(TRACKED_EVENTS.ACTIVATE);
```
jsines, line 16: Curious if it might make sense to track OS version here as well instead of adding a separate event for it? Seems like the best place for it 
https://git.hubteam.com/hubspot/cms-developer-tooling/issues/59

jsines: Oh, looks like we track that for every event. A separate tracker may be redundant then 😄

jsines (approved):
Small comments otherwise LGTM 👍

package.json
miketalley: One thing that I'm worries about is if someone is using the extension capabilities outside of the config and wants to disable tracking 🤔

src/lib/commands/config.ts
miketalley: Thanks for clarifying, that makes total sense. I don't think it really matters, though I do think that `string | undefined` is a bit more explicit and lends itself to later updating to something like `InputBoxResponseType` or something like that. 🤷

PR #155, opened 2022-12-08
Create serverless function folder

jsines:
- Moves create new module to a new "Create" submenu

- Adds "New functions folder" option under "Create" submenu 
Screen Shot 2022-12-08 at 4 43 34 PM
Screen Shot 2022-12-08 at 4 44 19 PM

The default `serverless.json`:
```
{
  "runtime": "nodejs18.x",
  "version": "1.0",
  "environment": {},
  "secrets": [],
  "endpoints": {
    "example": {
      "method": "GET",
      "file": "example.js"
    }
  }
}
```
And `example.js` (this name can easily be changed if we want something else, `serverless.js` perhaps?
```
// Require axios library to make API requests
const axios = require('axios');

// This function is executed when a request is made to the endpoint associated with this file in the serverless.json file
exports.main = ({ accountId }, sendResponse) => {
  // Use axios to make a GET request to the search API
  axios
    .get('https://api.hubapi.com/contentsearch/v2/search', {
      params: {
        portalId: accountId,
        term: 'searchTerm',
      },
    })
    .then(function(response) {
      // Handle success
      // The console.log statement will appear in the terminal when you run the hs logs CLI command
      // For full documentation, see: developers.hubspot.com/docs/cms/developer-reference/local-development-cms-cli#logs
      console.log('Data received from the search API:', response.data);
      // sendResponse is what you will send back to services hitting your serverless function
      sendResponse({ body: { response: response.data }, statusCode: 200 });
    })
    .catch(function(error) {
      // Handle error

      // This is a simple example; error handling typically will be more complicated.
      // For more information on error handling with axios, see: https://github.com/axios/axios#handling-errors
      sendResponse({ body: { error: error.message }, statusCode: 500 });
    });
};
```
Both are generated via `createFunction` in `@hubspot/cli-lib`


src/lib/fileHelpers.ts
```diff
@@ -61,13 +75,78 @@ export const convertFolderToModule = (
...
+    );
+  };
+};
+
+/*
+ * Returns a non-duplicate folder name with the format <folderName>.functions
+ * It will add a number to the end of the folder name if it already exists
+ * @param folderName - name of a folder with or without a .functions extension
+ */
+const getUniqueFunctionsFolderName = (folderPath: string) => {
```
jsines, line 97: This function (`getUniqueFunctionsFolderName`) and `convertFolderToServerlessFunction` are super similar to the module folder generation code, to the extent that we could generalize and make it a little DRYer, but I'm not sure it would make the code any more readable and these should be the only two cases we need to use this logic so not a ton of benefit to it.

jsines: Actually changed my mind on this overnight, I think `getUniqueFunctionsFolderName` and `getUniqueModulesFolderName` should be generalized at least, other two are ok to have separate.

src/lib/commands/serverlessFunction.ts
```diff
@@ -0,0 +1,19 @@
+import * as vscode from 'vscode';
+import { convertFolderToServerlessFunction } from '../fileHelpers';
+import { COMMANDS } from '../constants';
+
+export const registerCommands = (context: vscode.ExtensionContext) => {
+  context.subscriptions.push(
+    vscode.commands.registerCommand(
+      COMMANDS.CREATE_SERVERLESS_FUNCTION_FOLDER,
+      async (clickContext) => {
```
anthmatic, line 9: Does this need to be async?

TanyaScales:
> And `example.js` (this name can easily be changed if we want something else, `serverless.js` perhaps?

Yeah, I think `serverless.js` is better, IMO. Just to keep things in line with the intention of the overarching folder. I expect many developers will rename this to something unique anyways.

src/lib/commands/serverlessFunction.ts
jsines: Looks like no! @miketalley had this on his original module create code and I blindly copied it 😶 removed!

TanyaScales (approved):
LGTM! Tested out in the UI as well and it's working great

PR #162, opened 2022-12-12
Fix bug when resolving shallow relative paths

anthmatic:
When referencing a file within the same directory by relative path, the extension was incorrectly presenting an error that the file could not be found:
image

If you notice, the file BE assumes that this path is at the root level because we're not passing a `template_path` value along with the file source.

This PR adds that `template_path`, so all file paths returned from the BE are not absolute paths based on the user's local FS. This doesn't affect any linting on the BE and let's the extension do a proper check to see if the file exists:

image

image

### DIFF
--- a/src/lib/validation.ts
+++ b/src/lib/validation.ts
@@ -47,8 +47,7 @@ const isFileInWorkspace = (error: any) => {
   if (error.category === 'MODULE_NOT_FOUND') {
     filePath = filePath + '.module';
   }
-
-  return fs.existsSync(path.resolve(dirToActiveFile, filePath));
+  return fs.existsSync(path.normalize(filePath));
 };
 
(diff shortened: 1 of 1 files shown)


src/lib/validation.ts
```diff
@@ -47,8 +47,7 @@ const isFileInWorkspace = (error: any) => {
   if (error.category === 'MODULE_NOT_FOUND') {
     filePath = filePath + '.module';
   }
-
-  return fs.existsSync(path.resolve(dirToActiveFile, filePath));
+  return fs.existsSync(path.normalize(filePath));
```
anthmatic, line 50: `.normalize()` to make sure path is in correct local path format

TanyaScales (approved):

PR #164, opened 2022-12-13
Report issue link

miketalley:


src/lib/providers/helpAndFeedback.ts
```diff
@@ -18,6 +18,10 @@ export class HelpAndFeedbackProvider implements TreeDataProvider<any> {
         label: 'CLI Documentation',
         url: 'https://developers.hubspot.com/docs/cms/developer-reference/local-development-cli#interacting-with-the-developer-file-system',
       },
+      {
+        label: 'Report issue',
+        url: 'https://github.com/HubSpot/hubspot-cms-vscode/issues/new',
```
TanyaScales, line 23: This looks good to send to right now -- I think we'll want a pre-formatted "bug" template in the repo that asks for all the information we determine is required for triaging and solving bugs that come in from developers -- we can easily update the link when we create that template though. I'll start a comment on the original issue to talk about the template and what we would need in it.

TanyaScales (approved):

PR #167, opened 2022-12-16
Feedback

miketalley:
### Description
Adds a feedback form that can be used to submit experience rating info.

- Adds a feedback request command that shows an information message asking for feedback (opens feedback form panel)
  - Initially shown 3 days after install
  - Dismissed for 60 days if "Not now" or "X" clicked
  - Dismissed for 90 days if form is completed   
- Adds a link in Help & Feedback section to open the feedback form 
- Adds a feedback form that opens in a webview panel within VSCode
  - Submitting form sends the data to a [serverless function](https://app.hubspot.com/design-manager/23748177/file/97996466276) that saves it into a [hubdb table](https://app.hubspot.com/hubdb/23748177/table/5697923)

### Images
image

image

image


package.json
```diff
@@ -421,7 +437,9 @@
     },
     "dependencies": {
         "@hubspot/cli-lib": "^4.1.6-beta.0",
+        "axios": "^1.2.2",
```
anthmatic, line 440: Since this is just for the serverless function should this be externalized in the webpack build?

src/lib/panels/feedback.ts
```diff
@@ -0,0 +1,297 @@
...
+                <label>Visual Studio Code Version</label><br />
+                <input name="vs-code-version" type="text" value="${userIdentificationInformation.vscodeVersion}" disabled>
+            </div>
+            <div class="form-field hidden">
+                <label>HubSpot Extension Version</label><br />
+                <input name="hubspot-extension-version" type="text" value="${userIdentificationInformation.version}" disabled>
+            </div>
+            <div class="form-field hidden">
+                <label>Machine ID</label><br />
+                <input name="machine-id" type="hidden" value="${userIdentificationInformation.machineId}" disabled>
```
anthmatic, line 158: does this need to be disabled?

anthmatic: also wondering if we could use `readonly` instead of `disabled`, I think readonly still submits

package.json
miketalley: Since it's used in `src/lib/panels/feedback.ts` it needs to be a dep for the repo.

src/lib/panels/feedback.ts
miketalley: Good call, this works much better since I don't need the logic for enabling the disabled inputs.

package.json
anthmatic: ah, sorry I missed that! Out of curiosity, any reason not to use `fetch` on the VS Code side?

src/lib/panels/feedback.ts
```diff
@@ -0,0 +1,295 @@
...
+      'webview-ui',
+      'js',
+      'feedback-form.js',
+    ]);
+    const stylesheetUri = getUri(webview, extensionUri, [
+      'webview-ui',
+      'css',
+      'styles.css',
+    ]);
+    const userIdentificationInformation = getUserIdentificationInformation();
```
jsines, line 125: Should we avoid collecting this if the user has telemetry disabled? Since there's indication we're tracking that here users might expect to be anonymized.

src/lib/panels/feedback.ts
```diff
@@ -0,0 +1,295 @@
...
+              commands.executeCommand(
+                COMMANDS.GLOBAL_STATE.UPDATE_DELAY,
+                GLOBAL_STATE_KEYS.DISMISS_FEEDBACK_INFO_MESSAGE_UNTIL,
+                90,
+                'day'
+              );
+
+              trackEvent(TRACKED_EVENTS.FEEDBACK.FEEDBACK_PANEL_SUBMITTED);
+
+              window.showInformationMessage('Feedback submitted! Thanks you!');
```
jsines, line 273: Nit: 
```suggestion
              window.showInformationMessage('Feedback submitted! Thank you!');
```

serverless/vscodeExtensionFeedbackSubmit.functions/submit.js
```diff
@@ -0,0 +1,48 @@
...
+    method: 'POST',
+    url: `${HUBDB_API_PATH}/tables/${VSCODE_EXTENSION_FEEDBACK_HUBDB_TABLE_NAME}/draft/push-live`,
+    data: {},
+    headers: {
+      authorization: `Bearer ${VSCODE_EXTENSION_FEEDBACK_ACCESS_TOKEN}`,
+    },
+  });
+};
+
+exports.main = async (context, sendResponse) => {
```
jsines, line 32: With this being public especially, is there any concern about malicious behavior here? Is there any protection against someone spamming it with junk data? I'm assuming the the hubdb API already cleansing inputs, not familiar with what security serverless functions offer out of the box.

src/lib/panels/feedback.ts
miketalley: Updated `getUserIdentificationInformation` to respect telemetry settings.

serverless/vscodeExtensionFeedbackSubmit.functions/submit.js
miketalley: Yes, definitely is a concern, but I think that is an issue that we deal with elsewhere as well. The endpoint to add the row to hubdb should be able to handle strings that are too long, etc.

I added some `maxlength`s to the inputs. If you have any other ideas of what else can be done, happy to add.

package.json
anthmatic: Sorry for the "unresolve comment" just wanted to follow up to see if we could avoid adding this dep? We already have `node-fetch` which is used for snippet generation and vscode makes `fetch` available as well.

serverless/vscode-extension-feedback-hubdb-table.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "vscode_extension_feedback_form_submissions",
+  "useForPages": false,
+  "label": "Feedback Form Submissions",
+  "allowChildTables": false,
+  "allowPublicApiAccess": true,
```
anthmatic, line 6: Can we make this `false`?

package.json
miketalley: I had tried to use `node-fetch` before, but I'm running into an issue with using it. I think it may be because we are working within the panel, which is a webview, we aren't technically in node anymore.

miketalley: It was an import issue. I was able to remove `axios` in favor of `node-fetch`.

PR #168, opened 2022-12-16
Refactor create

### LINKED ISSUE #130: Create new file from template
Add a way to create new templates in a similar way to Design Manager

jsines:
Adds right click create options for serverless functions, templates, sections, partials, and global partials.

Refactors all right click create to use common logic and changes to `onDidCreateFiles` rather than `onWillCreateFiles`

Requires changes to cli-lib, see: https://github.com/HubSpot/hubspot-cli/pull/790

And an additional sample asset: https://github.com/HubSpot/cms-sample-assets/pull/1

Closes #130


src/lib/constants.ts
```diff
@@ -41,8 +41,15 @@ export const COMMANDS = {
...
-  CREATE_SERVERLESS_FUNCTION_FOLDER: 'hubspot.create.serverlessFunctionFolder',
+  CREATE: {
+    MODULE: 'hubspot.create.module',
+    SECTION: 'hubspot.create.section',
+    TEMPLATE: 'hubspot.create.template',
+    PARTIAL: 'hubspot.create.partial',
+    GLOBAL_PARTIAL: 'hubspot.create.globalPartial',
+    SERVERLESS_FUNCTION: 'hubspot.create.serverlessFunction',
+    SERVERLESS_FUNCTION_FOLDER: 'hubspot.create.serverlessFunctionFolder',
+  },
```
jsines, line 52: @miketalley cleaned this up a little, may require changes to tracking code (?)

src/lib/commands/create.ts
```diff
@@ -0,0 +1,182 @@
...
+
+export const registerCommands = (context: ExtensionContext) => {
+  context.subscriptions.push(
+    commands.registerCommand(
+      COMMANDS.CREATE.SECTION,
+      onClickCreateFile('html', (uniqueFilePath: string) => {
+        createTemplate(
+          path.basename(uniqueFilePath),
+          path.dirname(uniqueFilePath),
+          'section',
```
jsines, line 77: `'section'` here refers to `section.html` in https://github.com/HubSpot/hubspot-cli/tree/create-serverless-functions/packages/cli-lib/defaults

and likewise for other template creation commands

anthmatic: Could we create constants for these file types?

src/lib/commands/create.ts
```diff
@@ -0,0 +1,182 @@
...
+      doAfter(uniqueFilePath);
+    }
+  };
+};
+
+// doAfter receives (filepath: string)
+const onClickCreateFile = (extension: string, doAfter: Function) => {
+  return (clickContext: any) => {
+    if (clickContext.scheme === 'file') {
+      const createFileSubscription = workspace.onDidCreateFiles(
```
miketalley, line 47: There was an issue with module creation using `onDidCreateFiles` instead of `onWillCreateFiles` before. When creating a module folder using "MyModule" in the input, the "MyModule" folder would show up briefly, then get updated to "MyModule.module". Does switching to `onDidCreateFiles` cause a regression?

src/lib/constants.ts
miketalley: The tracking PR got merged in(https://github.com/HubSpot/hubspot-cms-vscode/pull/157). It looks like the tracked event for module creation would be removed in this PR. Can we add that one back? We should probably add tracked events for the other create commands as well.

src/lib/commands/create.ts
jsines: Yep, but this behavior seems unavoidable. `onWillCreateFiles` [doesn't allow edits to the file being created,](https://hubspot.slack.com/archives/C042U2U4NMC/p1671214164062939) and I'm not sure there's another way to do it while still keeping the name prompting in the explorer rather than a popup.

src/lib/constants.ts
jsines: Added 👍 

I put them at the end of the callback and omitted the `await` since it shouldn't impact the rest of the logic and just needs to be fired off, see any issue with that?

src/lib/commands/create.ts
```diff
@@ -0,0 +1,201 @@
...
+    commands.registerCommand(
+      COMMANDS.CREATE.PARTIAL,
+      onClickCreateFile('html', async (uniqueFilePath: string) => {
+        await createTemplate(
+          path.basename(uniqueFilePath),
+          path.dirname(uniqueFilePath),
+          TEMPLATE_NAMES.PARTIAL,
+          { allowExisting: true }
+        );
+        trackEvent(TRACKED_EVENTS.CREATE.PARTIAL);
```
anthmatic, line 110: I think `trackEvent` is async

jsines: It is async, but since it doesn't interfere with any logic and we don't do anything with the result isn't it OK to omit the `await`?

src/lib/commands/create.ts
miketalley: Do you mean edits to the file contents? What if we were to have `onClickCreateFolder` use `onWillCreateFiles` and `onClickCreateFile` use `onDidCreateFiles`?

miketalley (commented):
Looking good -- I think you may want to merge in master and make sure we didn't lose anything from Module creation fix (see https://github.com/HubSpot/hubspot-cms-vscode/pull/176/files#r1061682427).

Also, I commented on one possibility to fix the folder showing up prior to having the final name associated with it.

src/lib/commands/create.ts
jsines: We could definitely do that, I figured the additional complexity wasn't worth it for such a minor difference when we wouldn't be able to avoid the behavior for file creation but I'll add it back if you prefer it that way

jsines:
> I think you may want to merge in master and make sure we didn't lose anything from Module creation fix

Should be accounted for, I opted to keep `createModule` and `createTemplate` instead of directly downloading github contents - `await`ing them now since they're async.

PR #171, opened 2022-12-19
Updating relevant account info

miketalley:
- Updates icon for account from `account` to `star-full` when the account is the default, and removes "(default)" text
- Adds "[QA]" to account name if `env: 'qa'`
- Adds right click menu option for "Show personal access key info" that opens browser to PAK
- Improves tooltip hover text for accounts

image
image


src/lib/providers/accounts.ts
```diff
@@ -11,6 +11,23 @@ import { HubspotConfig, Portal } from '../types';
...
+};
+
+const getAccountIdentifiers = (portal: Portal) => {
+  let accountIdentifiers = '';
+
+  if (portal.env === 'qa') {
+    accountIdentifiers = '[QA]';
+  }
+
+  return accountIdentifiers;
```
jsines, line 28:
```suggestion
  return portal.env === 'qa' ? '[QA]' : '';
```

jsines (approved):

PR #175, opened 2022-12-20
Telemetry prompt

jsines:
Adds toast on initial load to inform user about data collection. 

Screen Shot 2022-12-20 at 3 18 16 PM


anthmatic (approved):
👍

PR #176, opened 2022-12-20
Module creation fix

miketalley:
The assets needed to populate a created module are not available as part of the webpack bundle.
`@hubspot/cli-lib/defaults/Sample.module/...`

This PR works alongside a PR to the CLI (https://github.com/HubSpot/hubspot-cli/pull/792) that will move assets to a new package and create helper functions to download files from that package on github.

/cc @anthmatic @jsines @TanyaScales

### DIFF
--- a/src/lib/fileHelpers.ts
+++ b/src/lib/fileHelpers.ts
@@ -3,7 +3,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 
 const { createFunction } = require('@hubspot/cli-lib/functions');
-const { downloadGitHubRepoContentsAtPath } = require('@hubspot/cli-lib/github');
+const { downloadGitHubRepoContents } = require('@hubspot/cli-lib/github');
 
 const copySampleFunctionFilesToFolder = (folderPath: string) => {
   const { dir, base } = path.parse(folderPath);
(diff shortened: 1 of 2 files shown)


src/lib/fileHelpers.ts
```diff
@@ -80,8 +68,12 @@ export const convertFolderToModule = (
...
             edit.renameFile(e.files[0], Uri.file(uniqueModulePath));
 
-            workspace.applyEdit(edit).then(() => {
-              copySampleModuleFilesToFolder(uniqueModulePath);
+            workspace.applyEdit(edit).then(async () => {
+              await downloadGitHubRepoContents(
+                'hubspot-cli',
+                'packages/cli-assets/cms/modules/Sample.module',
+                uniqueModulePath
+              );
```
jsines, line 76: Doesn't technically matter (same behavior) but curious on your thought process calling the download directly here as opposed to keeping `createModule` and keeping the implementation abstracted out / handled by cli-lib

jsines (approved):

src/lib/fileHelpers.ts
miketalley: I think we have the need to download these sample assets in both the cli and here, so it makes sense to me to have the shared logic kept in `@hubspot/cli-lib` -- especially since that is already a dependency of this repo.

miketalley: Also FYI, I think this change may cause a conflict with your branch `refactor-create`. @jsines

jsines: Yep - even if we use keep using `createModule` it'll have to be async, as will templates. I'll update mine so that once `cli-lib` changes are done it'll be good to go 👍

PR #178, opened 2023-01-04
Testing infra

jsines:
Tests can be run with `npm run test:e2e`, currently only contains tests for if VSCode opened and if the extension loaded into the Activity Bar.

I enabled `this.retries(3)` because the test was running before the extension loaded and `this.timeout(n)` wouldn't seem to fix it. It passes the second time. Not sure if there's a way to hide the retried test's failure from the report.

Screen Shot 2023-01-04 at 1 34 57 PM

WDIO docs: https://webdriver.io/docs/wdio-vscode-service/

~~Also the build is failing with:
`error @wdio/cli@8.1.2: The engine "node" is incompatible with this module. Expected version "^16.13 || >=18". Got "12.22.12"`
Could we just change the node version in build.yml? I see we're also mixing package managers in there so it's pretty out of date haha~~

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,3 +3,4 @@ package-lock.json
 package.json
 .vscode
 dist
+.wdio-vscode-service
--- a/test/specs/example.e2e.ts
+++ b/test/specs/example.e2e.ts
@@ -1,7 +1,6 @@
 describe('WDIO VSCode Service', function () {
-
   before(() => browser.pause(3000));
-  
+
   it('should be able to load VSCode', async () => {
     const workbench = await browser.getWorkbench();
     expect(await workbench.getTitleBar().getTitle()).toBe(
--- a/test/specs/example.e2e.ts
+++ b/test/specs/example.e2e.ts
@@ -1,6 +1,7 @@
 describe('WDIO VSCode Service', function () {
-  this.retries(3);
 
+  before(() => browser.pause(3000));
+  
   it('should be able to load VSCode', async () => {
     const workbench = await browser.getWorkbench();
     expect(await workbench.getTitleBar().getTitle()).toBe(
--- a/test/specs/example.e2e.ts
+++ b/test/specs/example.e2e.ts
@@ -1,0 +1,7 @@
+describe('WDIO VSCode Service', () => {
+    it('should be able to load VSCode', async () => {
+        const workbench = await browser.getWorkbench()
+        expect(await workbench.getTitleBar().getTitle())
+            .toBe('[Extension Development Host] Visual Studio Code')
+    })
+})
--- a/test/specs/example.e2e.ts
+++ b/test/specs/example.e2e.ts
@@ -1,11 +1,13 @@
-describe('WDIO VSCode Service', () => {
+describe('WDIO VSCode Service', function() {
+    this.retries(3);
+
     it('should be able to load VSCode', async () => {
         const workbench = await browser.getWorkbench()
         expect(await workbench.getTitleBar().getTitle())
             .toBe('[Extension Development Host] hubspot-cms-vscode')
     })
     
-    it('should show all activity bar items', async () => {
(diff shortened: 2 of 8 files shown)


test/specs/example.e2e.ts
```diff
@@ -0,0 +1,23 @@
...
+    const workbench = await browser.getWorkbench();
+    const viewControls = await workbench.getActivityBar().getViewControls();
+    expect(await Promise.all(viewControls.map((vc) => vc.getTitle()))).toEqual([
+      'Explorer',
+      'Search',
+      'Source Control',
+      'Run and Debug',
+      'Extensions',
+      'HubSpot',
+    ]);
```
miketalley, line 21: If someone was to change the order of their controls, would that break this?

miketalley (commented):
This is really cool!

I had an issue running it locally, I got an error...
```
[chrome 102.0.5005.167 mac os x #0-0] expect(received).toEqual(expected) // deep equality

- Expected  - 1
+ Received  + 0
```
... although it seems like the test passed in the end. Is this expected? Is there anything we can do to prevent this? Perhaps utilize the hooks that are provided by vscode?

Also, how did you decide on using `wdio` for this? Would you mind providing some more info about what you found when looking at the various options available for integrating testing?

test/specs/example.e2e.ts
```diff
@@ -0,0 +1,23 @@
+describe('WDIO VSCode Service', function () {
+  this.retries(3);
```
anthmatic, line 2: I think that we can do something like `before(() => browser.pause(3000));`

jsines:
> Is this expected? Is there anything we can do to prevent this? Perhaps utilize the hooks that are provided by vscode?

I have `this.retries(3)` in there because the test seems to run before the extension loads the first time, spent a bit of time trying to fix it through WDIO but would be a good idea to see if there's a way to do it through VSCode API 👀. I'll do a little more digging 

> Also, how did you decide on using wdio for this? Would you mind providing some more info about what you found when looking at the various options available for integrating testing?

I posted a bit about reasoning [here](https://hubspot.slack.com/archives/C042U2U4NMC/p1671641953736809). Also had considered selenium with https://www.npmjs.com/package/vscode-extension-tester since we use selenium for everything else but documentation seemed less extensive, and their docs explicitly said "MacOS support is limited" so I opted for WDIO instead.

anthmatic:
This looks great! 

I'm unsure of the scope for this. Do you plan to add coverage for some other parts of the extension or just handling the infra for now?

jsines:
> Do you plan to add coverage for some other parts of the extension or just handling the infra for now?

Just handling infra right now, we can prioritize as a team which functionality is most important to add tests for pre-launch & add e2e tests incrementally with new features.

test/specs/example.e2e.ts
jsines: That worked! Thank you 🙇‍♀️

test/specs/example.e2e.ts
jsines: Just tested on my machine, order on my local VSCode instance had no impact on the one `wdio` spins up

anthmatic:
Adding this mostly for the sake of documenting it: 

The `.wdio-vscode-service` ends up adding a ton of scripts to the VS Code "NPM Scripts" pane. I don't think there's much we can do to help that, but if you use this (as I sometimes do) you can add this to your settings:

  "npm.exclude": ["**/.wdio-vscode-service/**"]

anthmatic (approved):
Just one not to add to `vscodeignore`, otherwise works great!

.prettierignore
```diff
@@ -3,3 +3,4 @@ package-lock.json
 package.json
 .vscode
 dist
+.wdio-vscode-service
```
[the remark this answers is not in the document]
jsines: Good catch, we should add the `test` directory as well to avoid shipping that.

PR #189, opened 2023-01-19
[Do Not Merge] Run Tests during GH Action Build

anthmatic:
I figured it makes sense to regularly run the tests against new changes and this seemed like the best place but I'm open to other ideas!

I tried testing this locally with Act and it was not working but I believe that's because of the linux Docker image used. From what I can tell, GH Actions does support this but there might be some follow up testing needed after merging.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,3 +16,6 @@ jobs:
       - name: Lint
         run: |
           npm run lint
+      - name: E2E Tests
+        run: |
+          npm run test:e2e


test/specs/example.e2e.ts
```diff
@@ -3,8 +3,8 @@ describe('WDIO VSCode Service', function () {
 
   it('should be able to load VSCode', async () => {
     const workbench = await browser.getWorkbench();
-    expect(await workbench.getTitleBar().getTitle()).toBe(
-      '[Extension Development Host] hubspot-cms-vscode'
+    expect(await workbench.getTitleBar().getTitle()).toContain(
+      'hubspot-cms-vscode'
     );
```
anthmatic, line 8: Just making test a little more flexible since the titlebar seems to be different on different operating systems

PR #201, opened 2023-02-03
Prevent error if hs already exists in the user's bin

jsines:
If the old cms-cli was already installed on a user's machine, and the extension runs npm commands to install hubspot-cli, npm will crash in the process because it will not overwrite the existing executable without --force.

This change adds a command to remove the legacy CLI during the update process to prevent this error.

### DIFF
--- a/src/lib/commands/terminal.ts
+++ b/src/lib/commands/terminal.ts
@@ -42,7 +42,7 @@ export const registerCommands = (context: ExtensionContext) => {
       ).includes('@hubspot/cms-cli');
       if (hsLegacyInstalled) {
         const selection = await window.showWarningMessage(
-          'The legacy Hubspot CLI (@hubspot/cms-cli) must be removed to update. Continue?',
+          'The legacy Hubspot CLI (@hubspot/cms-cli) will be removed to update. Continue?',
           ...['Okay', 'Cancel']
         );
         if (!selection || selection === 'Cancel') {
--- a/src/lib/commands/terminal.ts
+++ b/src/lib/commands/terminal.ts
@@ -37,6 +37,20 @@ export const registerCommands = (context: ExtensionContext) => {
       );
       terminal.show();
 
+      const hsLegacyInstalled: boolean = (
+        await runTerminalCommand(`npm list -g`)
+      ).includes('@hubspot/cms-cli');
+      if (hsLegacyInstalled) {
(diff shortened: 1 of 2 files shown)


src/lib/commands/terminal.ts
```diff
@@ -51,7 +51,7 @@ export const registerCommands = (context: ExtensionContext) => {
       }, POLLING_INTERVALS.FAST);
 
       terminal.sendText(
-        "echo 'Updating the HubSpot CLI.' && npm i -g @hubspot/cli@latest && echo 'Update complete. You can now close this terminal window.'"
+        "echo 'Updating the HubSpot CLI.' && npm remove -g @hubspot/cms-cli && npm i -g @hubspot/cli@latest && echo 'Update complete. You can now close this terminal window.'"
```
bmatto, line 54: nit: `remove` is an alias for uninstall - is there a reason for `remove` over `uninstall`?

jsines: Nope just habit from a yarn user 😅 
```suggestion
        "echo 'Updating the HubSpot CLI.' && npm uninstall -g @hubspot/cms-cli && npm i -g @hubspot/cli@latest && echo 'Update complete. You can now close this terminal window.'"
```

jsines:
>I wonder if this is a little heavy handed. Maybe there's some reason a developer has the old version installed ... maybe we check for the old version and if it's there alert them to the conflict it creates and provide the command to remove it?

Can totally do something like this, my thinking was that it's the same project just with a different name so a user pressing upgrade would expect the legacy version to be removed but a UI warning of the conflict and option to abort makes sense to me

src/lib/commands/terminal.ts
```diff
@@ -37,6 +37,20 @@ export const registerCommands = (context: ExtensionContext) => {
       );
       terminal.show();
 
+      const hsLegacyInstalled: boolean = (
+        await runTerminalCommand(`npm list -g`)
+      ).includes('@hubspot/cms-cli');
+      if (hsLegacyInstalled) {
+        const selection = await window.showWarningMessage(
+          'The legacy Hubspot CLI (@hubspot/cms-cli) must be removed to update. Continue?',
```
bmatto, line 45: Can we say "will be removed"?

bmatto (approved):
LGTM - thanks @jsines !

PR #206, opened 2023-02-22
Remote file viewing

jsines:
Adds remote file view to the side bar

The remote file view shows a read-only view of the developer filesystem of the portal selected as default

Screenshot 2023-02-22 at 1 50 56 PM

The view will automatically refresh when the default portal is changed, or is deleted from the extension's account list, and can be manually refreshed like so. 

Screenshot 2023-02-22 at 1 52 25 PM


package.json
```diff
@@ -248,6 +248,10 @@
                 "command": "hubspot.account.viewPersonalAccessKey",
                 "title": "Show personal access key info",
                 "icon": "$(link-external)"
+            },
+            {
+                "command": "hubspot.remoteFs.refresh",
+                "title": "Refresh"
```
jsines, line 254: Adds refresh command

jsines, line 254: The extension manifest (`package.json`) contains the extension "contributions". Here we define the new refresh command, add an option for it in the "view/title" submenu for the remote fs (this is what adds it to the `...` button), add the `hubspot.treedata.remoteFs` view to our extension's sidebar, and add welcome content to display when the view is empty

src/lib/commands/config.ts
```diff
@@ -144,6 +145,7 @@ export const registerCommands = (context: ExtensionContext) => {
                 );
               }
               await trackEvent(TRACKED_EVENTS.DELETE_ACCOUNT);
+              commands.executeCommand('hubspot.remoteFs.refresh');
```
jsines, line 148: Executes a refresh when the default portal changes (above) or is deleted (below)

src/lib/providers.ts
```diff
@@ -1,21 +1,35 @@
...
 
   context.subscriptions.push(
     commands.registerCommand(COMMANDS.ACCOUNTS_REFRESH, () => {
       console.log(COMMANDS.ACCOUNTS_REFRESH);
       accountProvider.refresh();
     })
   );
+  context.subscriptions.push(
+    workspace.registerTextDocumentContentProvider(scheme, remoteFileProvider)
+  );
```
jsines, line 32: A `TextDocumentContentProvider` essentially tells VSCode how it should open items with different URIs. In this case we register our custom `remoteFileProvider` to the scheme 'hubl' so that we can use it to open items that begin `hubl:`

src/lib/providers.ts
```diff
@@ -28,6 +42,13 @@ const initializeTreeDataProviders = (context: ExtensionContext) => {
     TREE_DATA.HELP_AND_FEEDBACK,
     helpAndFeedbackProvider
   );
+  window.registerTreeDataProvider(TREE_DATA.REMOTE, remoteFsProvider);
+  context.subscriptions.push(
+    commands.registerCommand(COMMANDS.REMOTE_FS.REFRESH, () => {
+      console.log(COMMANDS.REMOTE_FS.REFRESH);
+      remoteFsProvider.refresh();
+    })
+  );
```
jsines, line 51: We also register our `TreeDataProvider` (the remote fs view) and the command to refresh the remote fs view, which calls `refresh()` on said provider

src/lib/providers/remoteFileProvider.ts
```diff
@@ -0,0 +1,26 @@
...
+      const file = await download(
+        getPortalId(),
+        `/${decodeURIComponent(filepath)}`
+      );
+      return file.source;
+    } catch (e) {
+      console.log(e);
+    }
+  }
+})();
```
jsines, line 26: Fetches the contents of the selected file path from the default portal

src/lib/providers/remoteFsProvider.ts
```diff
@@ -0,0 +1,92 @@
...
+
+function isPathFolder(path: string) {
+  const splitPath = path.split('/');
+  const fileOrFolderName = splitPath[splitPath.length - 1];
+  const splitName = fileOrFolderName.split('.');
+
+  return !(
+    splitName.length > 1 && FOLDER_DOT_EXTENSIONS.indexOf(splitName[1]) === -1
+  );
+}
```
jsines, line 26: The file endpoint tells us whether a given path is a folder or not, but only after we fetch its contents. We render it in the tree when it shows up as a child first, so we have to infer from its name whether it's a file or folder. For example our initial root call returns something like
```
{
  "id": "/",
  "name": "",
  "folder": true, <-- tells us that the root '/' is a folder
  "children": [
    "test_tabs.module", <-- we want to render these in the tree now, but won't get the above info until this gets expanded
    "Marketplace",
    "boilerplate-test",
    "growth-test",
    "practicum",
    "local-CmsDefaultSystemPages"
  ],
  "updatedAt": 1642781510812,
  "createdAt": 1642781510812,
  "archivedAt": null
}
```

src/lib/providers/remoteFsProvider.ts
```diff
@@ -0,0 +1,92 @@
...
+export class RemoteFsProvider implements TreeDataProvider<FileLink> {
+  private _onDidChangeTreeData: EventEmitter<
+    FileLink | undefined | null | void
+  > = new EventEmitter<FileLink | undefined | null | void>();
+  readonly onDidChangeTreeData: Event<void | FileLink | null | undefined> =
+    this._onDidChangeTreeData.event;
+
+  refresh(): void {
+    this._onDidChangeTreeData.fire();
+  }
```
jsines, line 36: This + above two declarations set up refreshing, when the refresh command gets fired this refills the view with getChildren()

src/lib/providers/remoteFsProvider.ts
```diff
@@ -0,0 +1,92 @@
...
+          q.icon,
+          TreeItemCollapsibleState.None,
+          Uri.parse(q.url)
+        )
+      : new RemoteFsTreeItem(
+          q.label,
+          q.icon,
+          TreeItemCollapsibleState.Collapsed
+        );
+  }
```
jsines, line 50: Think of `getTreeItem()` as a render function for the given FileLink

src/lib/providers/remoteFsProvider.ts
```diff
@@ -0,0 +1,92 @@
...
+            icon: 'symbol-folder',
+          }
+        : {
+            label: f,
+            url: parent ? `hubl:${parent.path}/${f}` : `hubl:${f}`,
+            icon: 'file-code',
+          };
+    });
+    return Promise.resolve(fileOrFolderList);
+  }
```
jsines, line 70: `getChildren()` gets called with no arguments initially to get the initial tree data (in this case the root directory of the remote fs) and then gets subsequently called with a parent argument when items go from `TreeItemCollapsibleState.Collapsed` to `TreeItemCollapsibleState.Expanded`

src/lib/providers/remoteFileProvider.ts
```diff
@@ -0,0 +1,26 @@
+import * as vscode from 'vscode';
+const { download } = require('@hubspot/cli-lib/api/fileMapper');
+const { getPortalId } = require('@hubspot/cli-lib');
+
+export const RemoteFileProvider = new (class
+  implements vscode.TextDocumentContentProvider
```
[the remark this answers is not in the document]
jsines: Yep we have `printWidth: 80` for prettier and this clocks in at 88. Changing the rule will including re-running it on the whole repo so gonna keep it as is for now 🥲

src/lib/providers/remoteFsProvider.ts
```diff
@@ -0,0 +1,92 @@
...
+import { FileLink, RemoteFsDirectory } from '../types';
+
+const {
+  getDirectoryContentsByPath,
+} = require('@hubspot/cli-lib/api/fileMapper');
+const { getPortalId } = require('@hubspot/cli-lib');
+const { FOLDER_DOT_EXTENSIONS } = require('@hubspot/cli-lib/lib/constants');
+
+function isPathFolder(path: string) {
+  const splitPath = path.split('/');
```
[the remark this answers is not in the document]
jsines: Nope, but it's fine if it doesn't. 
`"filename".split('/') === ['filename']`
`"foldername/filename".split('/') === ['foldername', 'filename']`
And then next line we fetch the last value in the array

src/lib/providers/remoteFsProvider.ts
```diff
@@ -0,0 +1,92 @@
...
+  private _onDidChangeTreeData: EventEmitter<
+    FileLink | undefined | null | void
+  > = new EventEmitter<FileLink | undefined | null | void>();
+  readonly onDidChangeTreeData: Event<void | FileLink | null | undefined> =
+    this._onDidChangeTreeData.event;
+
+  refresh(): void {
+    this._onDidChangeTreeData.fire();
+  }
+  getTreeItem(q: FileLink): TreeItem {
```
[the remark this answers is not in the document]
jsines: Yes! Anthony originally had just `q` while he was working on it and status quo bias kept it like that in my head, `fileLink` is much better

src/lib/providers/remoteFileProvider.ts
```diff
@@ -0,0 +1,26 @@
+import * as vscode from 'vscode';
```
bmatto, line 1: NIT: prefer named imports i.e. `import { TextDocumentContentProvider } from 'vscode'`

bmatto (approved):
lgtm

PR #208, opened 2023-02-28
Win32 fixes

jsines:
Updates terminal usage and paths for win32 compatibility. 

Also updates tracking to be non-blocking, no longer `await`s it and instead just reports an error when the promise finishes if it fails.

### DIFF
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -34,21 +34,19 @@ export const runTerminalCommand = async (
 ): Promise<string> => {
   return new Promise((resolve, reject) => {
     try {
-      const cmd = process.platform === 'win32'
-        ? `${terminalCommand} & exit`
-        : `${terminalCommand} && exit`;
-      exec(
-        cmd,
-        (error: Error, stdout: string, stderr: string) => {
-          const err = error || stderr;
+      const cmd =
+        process.platform === 'win32'
+          ? `${terminalCommand} & exit`
+          : `${terminalCommand} && exit`;
+      exec(cmd, (error: Error, stdout: string, stderr: string) => {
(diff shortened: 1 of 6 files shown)


src/lib/helpers.ts
```diff
@@ -9,7 +9,7 @@ export const getRootPath = () => {
   if (!workspaceFolders || workspaceFolders.length < 1) {
     return;
   }
-  return workspaceFolders[0].uri.path;
+  return workspaceFolders[0].uri.fsPath;
```
bran-van-d, line 12: Does this adjust the folder issue when loading the extension with no folder selected/opened?

bran-van-d (approved):
Looks good to me!

src/lib/helpers.ts
jsines: Nope, that one's a little tricky because it has to do with how config gets set, we initially try to set config through the CLI via `getAccountId()` but that finds it via `findup` on `process.cwd()`, which in this case the extension lives in the VSCode folder so unless the user has their `hubspot.config.yml` in their system's root directory it won't find one. We fetch the root path that VSCode has open (since that's the most likely place for the intended config to be) and set it from there. There's probably a solution with switching around the order of events / passing cli-lib the correct path some other way but it's a fairly niche issue so I don't want to sink a ton of time into it.

This particular line is fixing a win32 only issue, on mac we have the same form
```
uri.path === /Users/jsines/src/test
uri.fsPath === /Users/jsines/src/test
```
but on win32 they look different
```
uri.path === /c:/Users/jsines/src/test
uri.fsPath === c:\\Users\\jsines\\src\\test
```
and the former has trouble being used in other areas.

PR #211, opened 2023-03-06
Remote FS deletion

jsines:
Adds delete functionality on right click for remote FS.

Screenshot 2023-03-06 at 2 00 28 PM

Includes a check to make sure the action was intentional.

Screenshot 2023-03-06 at 2 00 42 PM

### DIFF
--- a/src/lib/commands/remoteFs.ts
+++ b/src/lib/commands/remoteFs.ts
@@ -4,20 +4,25 @@ const { deleteFile } = require('@hubspot/cli-lib/api/fileMapper');
 const { getPortalId } = require('@hubspot/cli-lib');
 
 export const registerCommands = (context: ExtensionContext) => {
-    context.subscriptions.push(
-        commands.registerCommand(COMMANDS.REMOTE_FS.DELETE, async (clickedFileLink) => {
-            console.log(COMMANDS.REMOTE_FS.DELETE);
-            const fileLinkIsFolder = clickedFileLink.icon === 'symbol-folder';
-            const filePath = fileLinkIsFolder ? clickedFileLink.path : clickedFileLink.url.slice(5)
-            const selection = await window.showWarningMessage(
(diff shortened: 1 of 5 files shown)


src/lib/commands/remoteFs.ts
```diff
@@ -0,0 +1,28 @@
...
+export const registerCommands = (context: ExtensionContext) => {
+  context.subscriptions.push(
+    commands.registerCommand(
+      COMMANDS.REMOTE_FS.DELETE,
+      async (clickedFileLink) => {
+        console.log(COMMANDS.REMOTE_FS.DELETE);
+        const fileLinkIsFolder = clickedFileLink.icon === 'symbol-folder';
+        const filePath = fileLinkIsFolder
+          ? clickedFileLink.path
+          : clickedFileLink.url.slice(5);
```
jsines, line 15: Slice here because if it's not a folder its url will have the prefix "hubl:" to make it openable by the `remoteFileProvider`

src/lib/commands/remoteFs.ts
```diff
@@ -0,0 +1,28 @@
...
+        const selection = await window.showWarningMessage(
+          `Are you sure you want to delete ${filePath}? This action cannot be undone.`,
+          ...['Okay', 'Cancel']
```
jasonnrosa, line 18: Just poked around in the app to see what other tools have for the confirmation content (was mainly curious but will toss it down below if helpful at all). Some other examples include:
- Delete this file/folder (when deleting something from design manager)
- Delete (when deleting a page)
- Yes, delete (when deleting a blog)

jasonnrosa (approved):

src/lib/commands/remoteFs.ts
jsines: Gotcha! We have the `hs remove` command as well in the CLI that has no warning beforehand so I think this is probably fine enough warning.

PR #214, opened 2023-03-23
Remote FS upload

jsines:
Depends on https://github.com/HubSpot/hubspot-cms-vscode/pull/213/files (changes from that PR are included here)

Adds remote fs upload functionality. Doesn't include support for fieldsjs, I wanted to make this as simple as possible for now and I'm not sure that feature is fully supported yet / gets a ton of use, if we want to add it down the line there's certainly room for it but this is a pretty big chunk as-is.

### DIFF
--- a/src/lib/commands/remoteFs.ts
+++ b/src/lib/commands/remoteFs.ts
@@ -94,9 +94,9 @@ export const registerCommands = (context: ExtensionContext) => {
     commands.registerCommand(
       COMMANDS.REMOTE_FS.UPLOAD,
       async (clickedFileLink) => {
-        let srcPath;
+        let srcPath: string;
         if (clickedFileLink === undefined || !(clickedFileLink instanceof Uri)) { // check if Uri, because having a remoteFs tree item selected will show up here too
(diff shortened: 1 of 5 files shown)


src/lib/commands/remoteFs.ts
```diff
@@ -75,9 +82,142 @@ export const registerCommands = (context: ExtensionContext) => {
...
-        commands.executeCommand('hubspot.remoteFs.refresh');
+        deleteFile(getPortalId(), filePath).then(() => {
+          window.showInformationMessage(`Successfully deleted ${filePath}`);
+          invalidateParentDirectoryCache(filePath);
+          commands.executeCommand('hubspot.remoteFs.refresh');
+        });
+      }
+    )
+  );
+  context.subscriptions.push(
```
bmatto, line 93: I think I'm following the logic branching here - however the cyclomatic complexity is a bit high. Would it be easier to understand/maintain if we could break out some of these branches in their own functions? That would also allow the function name to help describe what the particular logic branch aims to do.

PR #215, opened 2023-03-28
Remote FS Watch

jsines:
Adds watch functionality to VSCE. Similar to upload, watch can be initiated either via right-click on the desired folder in the explorer, or via the dropdown context in the remote FS view. The user will then be prompted for the remote path with which to sync.

The synced folder is then given a special icon in the remote FS view, and the watch session can be cancelled with a right click action, or by initiating another session. Only one watch session can be active at a time.

Screenshot 2023-03-28 at 5 02 17 PM

Changes to the local filesystem are watched using `chokidar` in `cli-lib`, and a special event is added for VSCE to invalidate the remote FS cache when files are removed, added, or moved.

### DIFF
--- a/package.json
+++ b/package.json
@@ -262,8 +262,20 @@
                 "title": "Invalidate Cache"
             },
             {
+                "command": "hubspot.remoteFs.startWatch",
+                "title": "Start Watch"
+            },
+            {
+                "command": "hubspot.remoteFs.endWatch",
+                "title": "End Watch"
+            },
+            {
                 "command": "hubspot.remoteFs.upload",
                 "title": "Upload"
+            },
+            {
+                "command": "hubspot.remoteFs.watch",
+                "title": "Watch"
             },
             {
                 "command": "hubspot.remoteFs.fetch",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -428,6 +452,10 @@
                 {
                     "command": "hubspot.remoteFs.invalidateCache",
                     "when": "false"
+                },
+                {
+                    "command": "hubspot.remoteFs.startWatch",
+                    "when": "false"
                 }
```
jsines, line 459: `hubspot.remoteFs.watch` and `hubspot.remoteFs.endWatch` are UI hooks, `hubspot.remoteFs.startWatch` is an internal command. It might be worth finding a naming scheme that makes this obvious for maintainability's sake

src/lib/commands/remoteFs.ts
```diff
@@ -155,14 +202,6 @@ const getUploadableFileList = async (src: any) => {
...
 };
 
-const invalidateParentDirectoryCache = (filePath: string) => {
-  let parentDirectory = dirname(filePath);
-  if (parentDirectory === '.') {
-    parentDirectory = '/';
-  }
-  commands.executeCommand('hubspot.remoteFs.invalidateCache', parentDirectory);
-};
-
```
jsines, line 165: moved to `../helpers.ts`

bran-van-d (approved):
Tested this out locally and it looks good -- thanks for clearing up my questions Jessica ⭐

PR #216, opened 2023-04-04
Improve UX feedback during large uploads

jsines:
Fixes an issue where during a large upload no UX feedback occurred, leading to confusion for users. Now when uploading a folder via Upload, or the initial upload via Watch

- A `window.showInformationMessage` is displayed to bring attention to the upload initialization

- A status bar item is added until the upload process finishes

- A final `window.showInformationMessage` displays once the upload is finished.

This change requires a minor change to `cli-lib` to add an optional callback to `watch` to run after the initial upload completes. Here we use said callback to display upload messaging as well as invalidate the cache and refresh the display.

### DIFF
--- a/src/lib/providers/remoteFsProvider.ts
+++ b/src/lib/providers/remoteFsProvider.ts
@@ -6,10 +6,14 @@ import {
   ThemeIcon,
   EventEmitter,
   Event,
+  window,
 } from 'vscode';
 import { FileLink, RemoteFsDirectory } from '../types';
 import * as path from 'path';
-import { invalidateParentDirectoryCache } from '../helpers';
+import {
+  buildUploadingStatusBarItem,
+  invalidateParentDirectoryCache,
+} from '../helpers';
 const {
   getDirectoryContentsByPath,
 } = require('@hubspot/cli-lib/api/fileMapper');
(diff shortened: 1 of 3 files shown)


src/lib/providers/remoteFsProvider.ts
```diff
@@ -94,14 +98,29 @@ export class RemoteFsProvider implements TreeDataProvider<FileLink> {
...
+          notify: false,
+          commandOptions: {},
+          filePaths: filesToUpload,
+        },
+        () => {
+          uploadingStatus.dispose();
+          window.showInformationMessage(
+            `Initial upload complete! Now watching for changes.`
+          );
+          invalidateParentDirectoryCache(destPath);
```
bran-van-d, line 121: Why do we require this addition here?

bran-van-d (approved):
Looks good to me, just had one question!

src/lib/providers/remoteFsProvider.ts
jsines: With the linked change to [cli-lib](https://github.com/HubSpot/hubspot-cli/pull/825) the `watch` function now has an optional argument for a callback to run after the initial upload completes. This is said callback! 

This allows for a little more UX feedback around the initial upload for the watch command, and refreshing the remote fs view once it finishes. Since its async, without the additional hook the only time we can run the cache invalidation and refresh the view would be immediately, which would almost definitely before the upload has finished, leading to inconsistent remote fs states and causing user confusion.

PR #217, opened 2023-04-04
Clean up commands to consts, add tracking, refresh remotefs on default account change

jsines:
See title! Some of these `hubspot.remoteFs.refresh`s should be hard refreshes, they were just put in before there was a cache so there was no distinction, went ahead and switched them over. Also removed some unused imports and added tracking for basic remoteFs events.


bran-van-d (approved):
Nice updates -- do these various commands get tracked in Amplitude or somewhere else?

PR #220, opened 2023-04-14
Add @hubspot folder to remote FS view

jsines:
Adds @hubspot folder to the remote FS view. Always appears at the top of the root directory with a locked icon. Contents are split between themes and modules and sorted, the same way they appear in the Design Manager.

Screenshot 2023-04-14 at 3 46 51 PM

Also did some refactoring of the tree building code to hopefully make it a little easier to read. FileLinks are now just the logical aspect of the file / path and its properties, and all decorative / presentation stuff is in the RemoteFsTreeItem constructor. Feels much nicer that way!

### DIFF
--- a/src/lib/providers/remoteFsProvider.ts
+++ b/src/lib/providers/remoteFsProvider.ts
@@ -167,6 +167,11 @@ export class RemoteFsProvider implements TreeDataProvider<FileLink> {
       // Default content wasn't originally in this endpoint and so doesn't show up unless manually queried
       if (remoteDirectory === '/') {
         directoryContents.children = ['@hubspot', ...directoryContents.children];
+      } else if (remoteDirectory === '@hubspot') {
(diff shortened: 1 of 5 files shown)


src/lib/providers/remoteFsProvider.ts
```diff
@@ -175,53 +164,84 @@ export class RemoteFsProvider implements TreeDataProvider<FileLink> {
...
+          .filter((f: string) => !f.endsWith('.module'))
+          .sort();
+        const onlyModules = directoryContents.children
+          .filter((f: string) => f.endsWith('.module'))
+          .sort();
+        directoryContents.children = [...noModules, ...onlyModules];
+      }
       this.remoteFsCache.set(remoteDirectory, directoryContents);
     }
+    const buildFileLinkFromPath = this.curriedFileLinkFromPathBuilder(parent);
```
bmatto, line 185: ❤️

PR #222, opened 2023-04-19
Minor copy and rfs display changes

jsines:
Changes RFS view name to better signify that it belongs to the account selected as default

Screenshot 2023-04-19 at 2 06 46 PM

Adds a header to remote files when viewed to better signify that they're read-only (Thanks for the suggestion bmatto!) 

Screenshot 2023-04-19 at 2 07 43 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -505,7 +505,7 @@
                 },
                 {
                     "id": "hubspot.treedata.remoteFs",
-                    "name": "Remote File System"
+                    "name": "Remote File System (Default Account)"
                 },
                 {
                     "id": "hubspot.treedata.helpAndFeedback",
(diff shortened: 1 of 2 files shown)


bran-van-d (approved):
🌟

jazzyclimber:
Nit: Is it possible to add "read only" comment in there? I feel like that red path still helps. But I think folks still might be confused, even though there is that pop up on hover/when you try to edit the file (I forget when it shows up).

Other than that looks good!

jsines:
>Nit: Is it possible to add "read only" comment in there? I feel like that red path still helps. But I think folks still might be confused, even though there is that pop up on hover/when you try to edit the file (I forget when it shows up).

Totally! We can do `[[ READONLY: @remote/${filepath} ]]`

PR #223, opened 2023-04-21
Logger override, add error catch for delete 

jsines:
Now overrides the [cli-lib logger ](https://github.com/HubSpot/hubspot-cli/blob/8aafada9270602e912d89fc6a513816f5f7232bb/packages/cli-lib/logger.js#L32) to instead surface the error messages within the VSCode window. Also adds a catch/finally to delete function to not break if an error occurs.


jasonnrosa (approved):
Really great addition! ❤️

PR #226, opened 2023-05-03
Fix lang bug 

jsines:
This has been a bug for a super long time! Because `cli-lib` doesn't have a build step / use webpack, it fetches lang data from `en.lyaml` directly from the filesystem at runtime. Because we use a build step / webpack for the VSCode extension, we pack `cli-lib` and at runtime it fetches lang data with a relpath from an unexpected location! This was OK but not desirable before when this just showed up in logs, but now we have the `cli-lib` logger overridden and these errors surface in the VSCE UI. 

To fix this I've copied the contents of `en.lyaml` to the extension, and written a webpack loader (ones I could find online were either far outdated or used a different yaml loader which didn't work) to get language data built into the extension and then set it manually in `cli-lib`. This requires a minor `cli-lib` change: https://github.com/HubSpot/hubspot-cli/pull/836

To keep this copy up to date I've also added husky and a precommit step to compare it with the `en.lyaml` for the resolved version of `cli-lib` on the public repo. If it's different it'll automatically download it and add it to the commit. This should run totally automatically and require no manual upkeep! 🎉


scripts/sanityCheckEnLyaml.js
```diff
@@ -0,0 +1,20 @@
...
+
+console.log('Running precommit check for lang files...');
+(async () => {
+  const res = await fetch(
+    `https://raw.githubusercontent.com/HubSpot/hubspot-cli/v${cliLibVersion}/packages/cli-lib/lang/en.lyaml`
+  );
+  const txt = await res.text();
+  const localLyaml = fs.readFileSync('./src/lang/en.lyaml', 'utf8');
+  if (txt !== localLyaml) {
+    console.error('Local lyaml needs to be updated, replacing...');
```
bran-van-d, line 15: Can we update this message to include the reason why we need to update? i.e "Local lyaml out of sync with remote, updating...", "Local lyaml needs to be updated  because it is out of sync, replacing...".

bran-van-d (approved):
nice fix! Curious: what was the content of the error that the user was seeing the editor?

jsines:
>nice fix! Curious: what was the content of the error that the user was seeing the editor?
In the dev console this shows up as 

```[ERROR] Error loading language data:  Error: ENOENT: no such file or directory, open '/Users/jsines/src/hubspot-cms-vscode/lang/en.lyaml'```

as `cli-lib` tries to resolve `../lang/en.lyaml` from the `dist` folder after compilation.

At runtime when `i18n` tries to run without language data loaded this error occurs in the user's window:
  
Screenshot 2023-05-05 at 10 42 16 AM

Issue #70, opened 2020-09-11
Add icon for HUBL files


anthmatic:
Currently, when you switch to a HUBL variant of a language, the file icon changes:
![image](https://user-images.githubusercontent.com/9009552/92950059-1832af00-f42a-11ea-8c2f-07f3dfa4cab3.png)

It should either be the same as the underlying file type or we should add some sort of HS themed one

gcorne:
FWIW, I think the file type icons for the main language (e.g. html or css) are probably fine to start.

anthmatic:
It looks like support for these icons lives here: https://github.com/jesseweed/seti-ui and icons are keyed off of the file extension, so it's not immediately clear to me why the icon changes when only the language changes. 

For extensions, there is no contribution point where you could define what icon a new language should have, but there is an open feature request for this here: https://github.com/Microsoft/vscode/issues/14662

j-malt:
VSCode uses the language setting (in this case, HubL+HTML) to determine the file icon, among other language features. This is provided in the `package.json`, in [contributes.languages](https://code.visualstudio.com/api/references/contribution-points#contributes.languages). Unfortunately there is no way to set the desired icon as "Whatever the user's HTML file icon is", since VSCode considers a HubL+HTML file a completely different language type from an HTML file. The two options we have would be to

1. Repackage the default VSCode Seti-UI icons in HS-CMS-VSCE, or
2. Provide our own language icons

In option 1, we run into inconsistent sizing between default HTML files and our HubL+HTML files, since there is no way to control icon sizing when defining language-icons. As well, if the user is using a different icon theme (like vscode-icons), there would be inconsistencies between default HTML and HubL+HTML files. 

As such, I'd like to work on getting our own icons made for the HubL+HTML and HubL+CSS files. This route much better UX, imo. 

For icon sizing, Seti-UI provides a [Sketch](https://www.sketch.com/workspace/9d4eb79c-f9cb-47d2-ad97-b62ea1eaa8ad/shares) file with a [template](https://github.com/jesseweed/seti-ui/blob/master/svgs.sketch), but the guideline seems to be a 32x32 SVG with an 18x18 icon. Though I experimented a bit myself and it seems that a 24x24 right-aligned HubSpot logo inside of the 32x32 canvas leads to the best looking results in VSCode.

PR #232, opened 2023-07-21
Add HubL icons

j-malt:
Adds some HubL icons in response to #70. Design is not final, but have attached some screenshots of how they look within VSCode. 

Dark mode VSCode:
image

Light mode VSCode:
image

cc/ @jsines  @TanyaScales


bmatto:
Did we decide against a `</>` or `<>` indicator for the HTML file?

j-malt:
> Did we decide against a `</>` or `<>` indicator for the HTML file?

Honestly, I couldn’t figure out how to make it read properly as such a small icon if I included the HTML logo. Seemed to clash too much with the HS logo.

TanyaScales:
Unfortunately, the color for the html icon is orange (in the native VSCode icon set) -- so finding another color for that icon would have to happen.. I also think we should have the icon to the bottom right of the logo and sitting above the hubspot logo... 

Screenshot 2023-07-24 at 9 40 51 AM
Screenshot 2023-07-24 at 9 41 12 AM

or -- conversely -- what if we left the main file icon as the prominent icon and the hubspot logo as the add on...

Screenshot 2023-07-24 at 9 44 29 AM
Screenshot 2023-07-24 at 9 44 35 AM

jsines:
>or -- conversely -- what if we left the main file icon as the prominent icon and the hubspot logo as the add on...

I like this approach a lot! 

Could also be good to get some designer opinions on these, maybe they could mock something up for us

bmatto:
I prefer the language icon being large with the small sprocket badge as well.

TanyaScales:
> Option 1 looks nice. Here is another shot at it w/ slightly modified colours.
> 
> Light: image
> 
> Dark: image

Lets go with this ^ for the time being -- we can revisit the icons if we are able to get some design resources

TanyaScales (approved):

PR #241, opened 2023-07-25
Removing wdio testing suite

TanyaScales:
## Context

Based on a conversation with @jsines -- This testing is non-functional as it stands right now. The third party package we were using `wdio` was causing issues and roadblocked the team. I'm removing the testing files and associated code in the extension as the intention is to revisit adding testing back in (and starting fresh). Will create an issue to track that work


jsines (commented):
Should also be good to remove deps: everything under `@wdio/`, `chromedriver`, `wdio-chromedriver-service`, `wdio-vscode-service`. And the `test:e2e` script can go too!

jsines (approved):

PR #160, opened 2022-12-09
Support code snippets from the BE

anthmatic:
I had originally planned to fix the extra empty file that was generate in #148 but I realized that a while back we added a `codeSnippet` property to the [hubldoc endpoint](https://api.hubspot.com/cos-rendering/v1/hubldoc). Unfortunately, as you can see here https://github.com/HubSpot/hubspot-cms-vscode/compare/ap/support-codeSnippets?expand=1#diff-1942d5c2aca105453d76b423494f061cfe94a22144fcfe87eb80aedc6b7fb67fR1 a lot of the snippets are incorrect. 

The main issue is that snippets like
```
{% gallery ${1:slides} ${2:loop_slides} ${3:num_seconds} ${4:show_pagination} ${5:sizing} ${6:auto_advance} ${7:transition} ${8:caption_position} ${9:display_mode} ${10:lightboxRows} %}"
```
are missing the argument syntax, so ` ${1:slides}` should look more like `slides="${1:slides}"`

I'll work on getting those updated before merging this in.

PR #243, opened 2023-07-27
Support codeSnippets from BE

TanyaScales:
## Description

This is just a redo of #160 in order to get rid of conflicts and issues with my machine. I double checked the JinjavaDocFactory work that Anthony did in January to make sure the snippets were corrected, and they were. We should be safe to integrate this new functionality into the autogeneration of HubL tags.


j-malt (approved):
Lgtm 👍

Issue #248, opened 2023-08-30
Unordered file structure in Remote file system


tristanblackburn-lewis:
**Describe the bug**
There is n odiscernible order of files in the remote file system

**To Reproduce**
Go to Remote Fiel Structure and view a subfolder

**Expected behavior**
Files should be listed in alphanumeric order

**Screenshots**
<img width="292" alt="Screenshot 2023-08-30 at 13 55 11" src="https://github.com/HubSpot/hubspot-cms-vscode/assets/50985018/911508ca-6706-4e22-9202-21412a5c9801">

** Environment Versions: **
- VSCode Version: 1.80.1
- HubSpot VSCode Extension version: v1.2.1
- OS: macOS Ventura 13.4.1

**Additional context**
It's driving me mental.

jsines:
I can empathize with how annoying this is! Looks like I missed it when I was implementing the feature and we sort `@hubspot` but not other directories.

I just submitted [a fix](https://github.com/HubSpot/hubspot-cms-vscode/pull/249) for this. We're avoiding shipping any changes during [INBOUND](https://www.inbound.com/) this week but I should be able to get it out early next week at the latest.

Thanks so much for reporting!

TanyaScales:
@tristanblackburn-lewis - the new version released today (1.2.2) has the fix to this issue for you if you want to update your extension. Thanks for filing!

PR #249, opened 2023-09-05
Sort remote directory alphabetically

jsines:
Resolves https://github.com/HubSpot/hubspot-cms-vscode/issues/248

Minor oversight! We sort default directory contents when we split between themes & modules like we do in DMUI. But we don't sort the rest of the remote directory contents. This fixes that! 

Waiting to ship this out until after INBOUND.


TanyaScales (approved):
Wonderful - thank you for the quick fix!

Issue #257, opened 2023-12-28
Inconsistent/misleading highlighting for `not` logical operator


evelynn-causemic:
When using the `not` operator within an expression test, it is typically highlighted in the same way as other logical operators like `is`, `and`, or `or`. This is true when it is written on its own, or when paired with `and` and `or`.

However, when paired with an `is` test in the form of an `is not` comparison, the `not` portion loses its operator highlighting — this incorrectly implies both that `is not` isn't a valid comparison operator, and that `not` *is* a valid comparison on its own.

## To reproduce
Write any logical comparison using an `is not` operator while syntax highlighting is enabled in any HTML+HubL or CSS+HubL file.

## Expected behavior
`not` should remain highlighted as an operator when used in an `is not` comparison.  

It is less clear whether `not` should be highlighted as an operator outside of `is not`, `or not`, and `and not` comparisons, since `not` isn't a valid comparison operator when used on its own. 

## Screenshots
This image shows a combination of correct and incorrect/misleading highlighting behavior.
![image](https://github.com/HubSpot/hubspot-cms-vscode/assets/107708603/525a726b-d737-4575-ab41-2995ae679ed0)

## Environment versions
- VSCode Version: 1.85.1
- HubSpot VSCode Extension Version: v1.2.2
- OS: MacOS Sonoma 14.1.1

PR #260, opened 2024-01-26
Fix `is not` syntax highlighting

j-malt:
Context: Issue #257 

The token `is not` was being incorrectly captured in our syntax. This PR updates the syntax regex to correctly capture this token. Here is a test snippet from the original issue, and the updated syntax highlighting:

```html
  <div>
    {%- if classes and (classes is string) -%}
        {% set classes = classes|split(" ") %}
    {%- elif (classes is not string) and not (classes is string) or not foo -%}
        {% set classes = classes|split(",") %}
    {%- elif (classes is not string) and (classes is not iterable) or not foo -%}
        {% set classes = [] %}
    {%- endif -%}
  </div>
```

### Before
Screen capture 2024-01-26 at 12 30@2x

### After
Screen capture 2024-01-26 at 12 30@2x

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ To enable Emmet on your `html-hubl` files, you can map `html-hubl` to `html` in 
 
 ### IntelliSense Suggestions
 
-If you are would like to get IntelliSense suggestions when in snippet placeholders, you will need to add the following to your user settings:
+If you would like to get IntelliSense suggestions when in snippet placeholders, you will need to add the following to your user settings:
 
 `"editor.suggest.snippetsPreventQuickSuggestions": false`
 
--- a/syntaxes/hubl.json
+++ b/syntaxes/hubl.json
@@ -107,6 +107,10 @@
           "match": "\\s*\\b(filter)\\s+([a-zA-Z_][a-zA-Z0-9_]*)\\b"
         },
         {
+          "match": "\\b(and|else|if|in|import|(is )?not|or|recursive|with(out)?\\s+context)\\b",
+          "name": "keyword.control.hubl"
+        },
+        {
           "captures": {
             "1": {
               "name": "keyword.control.hubl"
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -37,7 +37,7 @@ To enable Emmet on your `html-hubl` files, you can map `html-hubl` to `html` in
 
 ### IntelliSense Suggestions
 
-If you are would like to get IntelliSense suggestions when in snippet placeholders, you will need to add the following to your user settings:
+If you would like to get IntelliSense suggestions when in snippet placeholders, you will need to add the following to your user settings:
```
j-malt, line 40: Also fixes a typo

syntaxes/hubl.json
```diff
@@ -106,6 +106,10 @@
           },
           "match": "\\s*\\b(filter)\\s+([a-zA-Z_][a-zA-Z0-9_]*)\\b"
         },
+        {
+          "match": "\\b(and|else|if|in|import|(is )?not|or|recursive|with(out)?\\s+context)\\b",
+          "name": "keyword.control.hubl"
+        },
```
j-malt, line 112: I moved this up in the file so that it takes precedence over the `x is y` syntax highlighting, which was breaking the `x is not y` syntax highlight, since it was being captured as `x (is) not y` instead of `x (is not) y`. I've also added the optional group `(is )?` so it captures `is not` here.

syntaxes/hubl.json
```diff
@@ -115,7 +119,7 @@
               "name": "variable.other.hubl.test"
             }
           },
-          "match": "\\s*\\b(is)\\s+([a-zA-Z_][a-zA-Z0-9_]*)\\b"
+          "match": "\\s*\\b(is)(?!\\s+not)\\s+([a-zA-Z_][a-zA-Z0-9_]*)\\b"
```
j-malt, line 122: Added a non-capturing group for `is` so that this doesn't match `x is not y`, only `x is y`. We need to keep this group around because the alternative (adding `is` to the group above) would match independent `is` tokens incorrectly, so `x not is y` would get highlighted as an operator which seems undesirable (really we should error on that but outside the scope of this change).

TanyaScales (approved):
Awesome, thanks for getting this quick fix in

Issue #247, opened 2023-08-21
`hubspot-local-dev-lib` migration


jsines:
Starting a place to collect these moving parts before too many pile up and the migration becomes a tangled mess! This is meant to act as a growing list until the migration is ready to go through & we have the time to do it. 

- [ ] add language data bundling to `hubspot-local-dev-lib` and remove it from this repo
- [ ] add `hubspot-local-dev-lib` dependency and remove `cli-lib` dependency
- [ ] reroute imports to use `hubspot-local-dev-lib`
- [ ] Move some type definitions to use definitions from `hubspot-local-dev-lib` instead
  - [ ] [HublValidationError](https://github.com/HubSpot/hubspot-local-dev-lib/pull/30#discussion_r1300393444)

PR #264, opened 2024-05-09
Migrate VSCE to use local-dev-lib instead of cli-lib

jsines:
Hits the couple tasks here: https://github.com/HubSpot/hubspot-cms-vscode/issues/247
Depends on this PR from local-dev-lib: https://github.com/HubSpot/hubspot-local-dev-lib/pull/134

Diff size is mostly `package-lock.json` and file removals 😄


.husky/pre-commit
```diff
@@ -1,5 +1,2 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
-
-node ./scripts/sanityCheckEnLyaml.js
```
jsines, line 4: A little historical context on these lang changes - with `cli-lib` there was no bundling process and the `lyaml` files didn't get shipped with the package, so I had set it up to download the lang data on commit to keep it updated and then we could bundle it like this. No longer necessary with `local-dev-lib` bundling its own lang data so we don't see those failures

src/lib/types.ts
```diff
@@ -42,18 +42,6 @@ export function instanceOfCommand(object: any): object is Command {
...
-  message: string;
-  lineno: number;
-  startPosition: number;
-  categoryErrors: {
-    fullName?: string;
-    path?: string;
-  };
-  category: string;
-}
-
```
jsines, line 56: This type exists in `local-dev-lib` now, so changing our usage to that one and removing this. Theres some others I'm sure that we could swap to LDL types but this is just one I specifically noted

src/lib/uri.ts
```diff
@@ -48,12 +49,18 @@ const handleAuthRequest = async (authParams: URLSearchParams) => {
...
+    throw err;
+  }
+  const updatedConfig = await updateConfigWithAccessToken(
+    token,
     personalAccessKey,
-    name,
     env,
-  });
+    name
+  );
```
jsines, line 63: `updateConfigWithPersonalAccessKey` no longer exists in LDL so fetching token first and then using `updateConfigWithAccessToken`

j-malt (approved):
Looks good!

TanyaScales (approved):
LGTM

PR #280, opened 2025-05-02
Feature: Add JSON schema validation for projects

kemmerle:
## Description and Context
This PR adds schema validation for HubSpot project components' JSON configuration files. More information is available in the Canvas I posted in Slack yesterday (Message me for access). 

Relies on https://git.hubteam.com/HubSpot/project-parsing-lib/pull/20. I've created an experimental tag of PPL to ease testing. 

## Screenshots

Validation in the `app-hsmeta.json` file:

Screenshot 2025-05-02 at 10 12 39 AM

Validation in a `card-hsmeta.json` file:

Screenshot 2025-05-02 at 10 12 22 AM

## TESTING 

I would use the north star project to test out the schema validation. You can follow the instructions in the [contributing guide](https://github.com/HubSpot/hubspot-cms-vscode/blob/master/CONTRIBUTING.md) to run the project locally. 

## TODO

- [ ] Address feedback 
- [ ] Update and release PPL

## Who to Notify

@brandenrodgers @camden11 @joe-yeager @TanyaScales

### DIFF
--- a/package.json
+++ b/package.json
@@ -588,7 +588,7 @@
     },
     "dependencies": {
         "@hubspot/local-dev-lib": "^3.1.0",
-        "@hubspot/project-parsing-lib": "0.0.16-experimental.0",
+        "@hubspot/project-parsing-lib": "0.0.17-experimental.0",
         "dayjs": "^1.11.7",
         "debounce": "1.2.1",
         "findup-sync": "^5.0.0",
--- a/package.json
+++ b/package.json
@@ -588,7 +588,7 @@
     },
     "dependencies": {
         "@hubspot/local-dev-lib": "^3.1.0",
-        "@hubspot/project-parsing-lib": "0.0.18-experimental.0",
+        "@hubspot/project-parsing-lib": "0.0.19-experimental.0",
         "dayjs": "^1.11.7",
         "debounce": "1.2.1",
         "findup-sync": "^5.0.0",
--- a/package.json
+++ b/package.json
@@ -587,15 +587,19 @@
         ]
     },
     "dependencies": {
-        "@hubspot/local-dev-lib": "^1.13.1",
+        "@hubspot/local-dev-lib": "^3.1.0",
+        "@hubspot/project-parsing-lib": "0.1.8",
         "dayjs": "^1.11.7",
+        "debounce": "1.2.1",
         "findup-sync": "^5.0.0",
         "fs-extra": "^9.0.1",
         "husky": "^7.0.4",
+        "jsonc-parser": "^3.3.1",
         "semver": "^7.5.2",
         "set-value": ">=4.0.1"
     },
     "devDependencies": {
+        "@types/debounce": "^1.2.1",
         "@types/findup-sync": "^4.0.2",
         "@types/js-yaml": "^4.0.5",
         "@types/node": "^15.12.4",
--- a/package.json
+++ b/package.json
@@ -588,7 +588,7 @@
     },
     "dependencies": {
         "@hubspot/local-dev-lib": "^3.1.0",
-        "@hubspot/project-parsing-lib": "0.0.17-experimental.0",
+        "@hubspot/project-parsing-lib": "0.0.18-experimental.0",
         "dayjs": "^1.11.7",
         "debounce": "1.2.1",
         "findup-sync": "^5.0.0",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
     "name": "hubl",
     "displayName": "HubSpot",
     "description": "HubSpot CMS Hub support for VS Code",
-    "version": "1.3.2",
+    "version": "1.4.2",
     "publisher": "HubSpot",
     "engines": {
         "vscode": "^1.63.0"
--- a/package.json
+++ b/package.json
@@ -588,7 +588,7 @@
     },
     "dependencies": {
         "@hubspot/local-dev-lib": "^3.1.0",
-        "@hubspot/project-parsing-lib": "0.0.19-experimental.0",
+        "@hubspot/project-parsing-lib": "0.2.0",
         "dayjs": "^1.11.7",
         "debounce": "1.2.1",
         "findup-sync": "^5.0.0",
--- a/package.json
+++ b/package.json
@@ -588,7 +588,7 @@
     },
     "dependencies": {
         "@hubspot/local-dev-lib": "^3.1.0",
-        "@hubspot/project-parsing-lib": "0.1.8",
+        "@hubspot/project-parsing-lib": "0.0.16-experimental.0",
         "dayjs": "^1.11.7",
         "debounce": "1.2.1",
         "findup-sync": "^5.0.0",
--- a/src/lib/appValidation.ts
+++ b/src/lib/appValidation.ts
@@ -1,0 +1,435 @@
+import {
+  workspace,
+  Uri,
+  Diagnostic,
+  DiagnosticSeverity,
+  languages,
+  Range,
+  DiagnosticCollection,
+  ExtensionContext,
+  TextDocument,
+} from 'vscode';
+import {
+  getIntermediateRepresentationSchema,
+  mapToInternalType,
+  createAjvInstance,
+  validateUid,
+  errorMessages,
+  Ajv,
+  ValidateFunction,
+  AnySchema,
+  ErrorObject,
+} from '@hubspot/project-parsing-lib';
+import { getAccountId } from '@hubspot/local-dev-lib/config';
+import { dirname } from 'path';
+import * as jsonc from 'jsonc-parser';
+import { debounce } from 'debounce';
+
+const hsProjectJsonFilename = 'hsproject.json';
+const VALIDATION_DEBOUNCE_TIME = 250;
+
+const AjvErrorKeyword = {
+  AdditionalItems: 'additionalItems',
+  AdditionalProperties: 'additionalProperties',
+  Dependencies: 'dependencies',
+  Enum: 'enum',
+  ExclusiveMaximum: 'exclusiveMaximum',
+  ExclusiveMinimum: 'exclusiveMinimum',
+  Maximum: 'maximum',
+  MaxItems: 'maxItems',
+  MaxLength: 'maxLength',
+  MaxProperties: 'maxProperties',
+  Minimum: 'minimum',
+  MinItems: 'minItems',
+  MinLength: 'minLength',
+  MinProperties: 'minProperties',
+  MultipleOf: 'multipleOf',
+  OneOf: 'oneOf',
+  Pattern: 'pattern',
+  Required: 'required',
+  Type: 'type',
+};
+
+type SchemaAndValidator = { schema: AnySchema; validator?: ValidateFunction };
+type SchemaCacheEntry =
+  | {
+      byType: Record<
+        string, // internal type
+        SchemaAndValidator
+      >;
+    }
+  | { error: string };
+
+// A schema cache maps a platform version to a set of schemas and validators indexed by the internal type.
+// Validators are compiled on demand and then cached.
+type SchemaCache = Record<
+  string, // platform version
+  SchemaCacheEntry
+>;
+
+let diagnosticsCollection: DiagnosticCollection;
+let ajv: Ajv;
+let schemaCache: SchemaCache = {};
+
+export function initializeAppValidation(context: ExtensionContext): void {
+  diagnosticsCollection =
+    languages.createDiagnosticCollection('hubspot-app-schema');
+  context.subscriptions.push(diagnosticsCollection);
+
+  ajv = createAjvInstance();
+
+  workspace.onDidOpenTextDocument(validateDocument);
+  workspace.onDidChangeTextDocument(
+    debounce(
+      (event) => validateDocument(event.document),
+      VALIDATION_DEBOUNCE_TIME
+    )
+  );
+  workspace.onDidSaveTextDocument(validateDocument);
+  workspace.onDidCloseTextDocument((doc) =>
+    diagnosticsCollection.delete(doc.uri)
+  );
+  workspace.textDocuments.forEach(validateDocument);
+}
+
+async function validateDocument(document: TextDocument): Promise<void> {
+  try {
+    const diagnostics = await validateDocumentDiagnostics(document);
+    diagnosticsCollection.set(document.uri, diagnostics);
+  } catch (error) {
+    console.error('Error validating document:', error);
+  }
+}
+
+async function validateDocumentDiagnostics(
+  document: TextDocument
+): Promise<Diagnostic[]> {
+  if (!document.fileName.endsWith('-hsmeta.json')) return [];
+  const projectRoot = await findProjectRoot(document.uri);
+  if (!projectRoot) return [];
+  const projectVersion = await getProjectVersion(projectRoot);
+  if (!projectVersion) return [];
+  // Any file that fails the above checks is one that we don't handle
+  let metaFile: any; // The type is whatever JSON the user provided us with
+  try {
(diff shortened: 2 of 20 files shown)


src/lib/appValidation.ts
```diff
@@ -0,0 +1,435 @@
...
+  }
+  return validator;
+}
+
+// VSCode doesn't provide a dirname function, that's in the github:/microsoft/vscode-uri package
+function vscodeDirname(uri: Uri): Uri {
+  return uri.with({ path: dirname(uri.path) });
+}
+
+async function findProjectRoot(start: Uri): Promise<Uri | null> {
```
brandenrodgers, line 406: Could we use `vscode.workspace.findFiles` for this? I'm not very familiar with how it works or exactly what it does, but I see it being used [here](https://github.com/HubSpot/hubspot-cms-vscode/blob/2a4498c36039e8b5efde26dfa20c993dc6fa61cd/src/lib/providers/fileCompletion.ts#L7)

brandenrodgers: Yeah looks like there's no built-in way to do that. We could potentially use it to locate every `hsproject.json` file, and then filter that list down to just the paths where `currentDir.path.startsWith(potentialParentHsProjectJSON.path)` is true. But apparently findFiles will look through the entire workspace, which isn't as performant as this. I'm surprised they don't have a better built in way to do it.

What you have here seems like the best way to do it. Maybe we could make it a little easier to follow by turning the try/catch for `fs.stat` into a dedicated "doesFileExist()` util? There's an existing fileHelpers file that it could live in.

src/lib/appValidation.ts
```diff
@@ -0,0 +1,435 @@
...
+      const parent = vscodeDirname(currentDir);
+      if (parent.fsPath === currentDir.fsPath) {
+        return null;
+      }
+      currentDir = parent;
+    }
+  }
+}
+
+async function getProjectVersion(projectRoot: Uri): Promise<string | null> {
```
brandenrodgers, line 423: Would it make sense to set up some sort of file watcher to watch for edits to the hsproject.json file? We'll want to update all the schemas whenever the user edits the platform version. It probably won't happen often, but would be nice to have just in case.

src/lib/appValidation.ts
```diff
@@ -0,0 +1,435 @@
+import {
```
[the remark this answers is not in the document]
brandenrodgers: Not a blocker, but poking around the repo a little more I'm noticing how hard it is to keep track of everything that's happening. Maybe we could add a new folder called validation (or diagnostics?) for this work to live in? And in a follow up PR we could pull the existing validation.ts and lint.ts functionality into it and rename them to something that's not so generic. Looks like they only work on cms assets. Just trying to think of ways for us to help create some more clarity in here.

src/lib/appValidation.ts
kemmerle: This is cool, I asked `cursor` to generate a response to this comment based on some bullet points, and I think it did a good job:

```
The file watcher isn't necessary because we already handle platform version changes effectively in our current implementation. Here's why:

1. Every time we validate a document (which happens on open, change, and save), we:

- Find the project root and read the hsproject.json file
- Get the current platform version
- Use that version to fetch and validate against the correct schemas

2. The schema cache (schemaCache) is keyed by platform version, so if the platform version changes:

- The next validation will use the new version
- It will automatically fetch and cache the new schemas for that version
- Old version schemas will naturally be unused

3. This approach is actually more robust than a file watcher because:

- It ensures we always have the correct schemas for the current platform version
- It handles cases where the file might be changed outside of VS Code
- It's simpler and has fewer moving parts
- It's already integrated with our existing validation flow

So while a file watcher might seem like a good idea at first, our current implementation already handles platform version changes effectively without the additional complexity of watching for file changes. 
```

I'm going to resolve this conversation for now, please feel free to reopen if necessary!

brandenrodgers: Oh sweet, that's pretty cool. Seems good to me then. I was testing a case where I open a JSON file, then edit the platform version, and then go back into the JSON file again. It doesn't re-fetch the schemas in that case, but it does re-fetch them as soon as I save the JSON file, which I think is good enough (especially considering that it's simpler)

src/lib/projectConfigValidation.ts
```diff
@@ -0,0 +1,437 @@
...
+  if (!validator(metaFile['config'])) {
+    let diagnostics: Record<string, Diagnostic> = {};
+    for (const error of validator.errors || []) {
+      const { message, nodePath, onlyKeyNode } = customizeAjvError(error);
+      if (!message) {
+        continue;
+      }
+      // This is a duplicate error, skip it
+      if (
+        diagnosticDeduplicationKey(nodePath, error.keyword, error.params) in
```
brandenrodgers, line 208: Slight nit, but we call `diagnosticDeduplicationKey` three times in here. Maybe we can call it just once at the start?

TanyaScales:
Hi! Going to look at this as soon as I can -- also cc @joeyblake @tenas97 and @Aaronm14 for visibility on this change

src/lib/projectConfigValidation.ts
```diff
@@ -0,0 +1,439 @@
...
+  }
+  return [];
+}
+
+function customizeAjvError(error: ErrorObject): {
+  message: string | undefined;
+  nodePath: (string | number)[];
+  onlyKeyNode: boolean;
+} {
+  let nodePath = [
```
brandenrodgers, line 253: What does nodePath do? There's some logic in here that I'm not really following.

src/lib/projectConfigValidation.ts
```diff
@@ -0,0 +1,439 @@
...
+      ),
+    ];
+  }
+  const validator = await getCachedValidatorForExternalType(
+    projectRoot,
+    projectVersion,
+    accountId,
+    metaFile.type
+  );
+  if ('error' in validator) {
```
joe-yeager, line 196: I don't think we should use the in operator here since it doesn't handle nullable cases well

src/lib/projectConfigValidation.ts
```diff
@@ -0,0 +1,439 @@
...
+async function getCachedSchemasForAllTypes(
+  projectRoot: Uri,
+  platformVersion: string,
+  accountId: number
+): Promise<SchemaCacheEntry> {
+  // If schema cache has a schema for this version, return it
+  // If there is an error, we're going to return that too
+  if (schemaCache[platformVersion]) {
+    return schemaCache[platformVersion];
+  }
```
joe-yeager, line 352: What is the lifetime of this cache?  Is it as long as the editor is open?  Maybe we should consider timing items out of the cache

src/lib/projectConfigValidation.ts
```diff
@@ -0,0 +1,439 @@
...
+  let response: Record<string, AnySchema>;
```
joe-yeager, line 353: Nit: we are only using response in the `try` block, so we can make it a const and block scope it there.

src/lib/projectConfigValidation.ts
```diff
@@ -0,0 +1,439 @@
...
+  try {
+    response = await getIntermediateRepresentationSchema({
+      projectSourceDir: projectRoot.fsPath,
+      platformVersion,
+      accountId,
+    });
+    let byType: Record<string, SchemaAndValidator> = {};
```
joe-yeager, line 360: Nit: could we give this a slightly more descriptive name?  Also, I believe this could be a const since we aren't reassigning it.

src/lib/projectConfigValidation.ts
```diff
@@ -0,0 +1,439 @@
...
+    for (const [key, schema] of Object.entries(response)) {
+      byType[key] = { schema };
+    }
```
joe-yeager, line 363: Why are we reshaping this?  If I am understanding this correctly it is just taking an object that looks like:
```
{
  APPLICATION: { 
   ... all the schema fields
  }
}
```
and converting it to this:
```
{
  APPLICATION: {
    schema: {  ... all the schema fields }
  }
}
```

src/lib/projectConfigValidation.ts
```diff
@@ -0,0 +1,439 @@
...
+    };
+    schemaCache[platformVersion] = result;
+    return result;
+  } catch (error) {
+    const message = `${getFailedToFetchSchemasError()}. Reload your window to try again`;
+    console.error(message, error);
+    const result = {
+      error: message,
+    };
+    schemaCache[platformVersion] = result;
```
joe-yeager, line 375: Do we want to cache the error?  How long will this error stay cached?

src/lib/projectConfigValidation.ts
```diff
@@ -0,0 +1,439 @@
...
+  }
+  return validator;
+}
+
+// VSCode doesn't provide a dirname function, that's in the github:/microsoft/vscode-uri package
+function vscodeDirname(uri: Uri): Uri {
+  return uri.with({ path: dirname(uri.path) });
+}
+
+async function findProjectRoot(start: Uri): Promise<Uri | null> {
```
joe-yeager, line 412: Is this finding the project root or the path to the `hsproject.json`?

src/lib/projectConfigValidation.ts
```diff
@@ -0,0 +1,439 @@
...
+  let validator = schemaAndValidator.validator;
+  if (!validator) {
+    validator = ajv.compile(schemaAndValidator.schema);
+    schemaAndValidator.validator = validator;
+  }
+  return validator;
+}
+
+// VSCode doesn't provide a dirname function, that's in the github:/microsoft/vscode-uri package
+function vscodeDirname(uri: Uri): Uri {
```
joe-yeager, line 408: Nit: We could rename this to `dirname` and maybe move it to the `fileHelpers .ts`

src/lib/projectConfigValidation.ts
```diff
@@ -0,0 +1,439 @@
...
+  params: any
+): string {
+  return `${keyword}::${nodePath.join('/')}::${JSON.stringify(params)}`;
+}
+
+function getKeyNode(node: jsonc.Node): jsonc.Node | undefined {
+  return node.parent?.type === 'property' ? node.parent.children?.[0] : node;
+}
+
+async function getCachedSchemasForAllTypes(
```
joe-yeager, line 343: I think this function is a misnomer because it is getting getting the cached schema if it exists and falling back to fetching the schemas.  Maybe we can just call it `getAllSchemas` or something along those lines

src/lib/projectConfigValidation.ts
kemmerle: `nodePath` is used to track the path to a specific node in the JSON document being validated.

1. We start with `config` as the root path (since all validation errors are within the `config` object)
2. Take the `instancePath` from the AJV error (which is a JSON Pointer string like `/config/someProperty/0/otherProperty`)
3. Split it by `/` and removes the first empty element (with `slice(1)`)
4. Map each part to either:

* A number (if it matches `\d+`, meaning it's an array index)
* The original string (if it's a property name)

For example, if there's an error in `config.someProperty[0].otherProperty`, the `nodePath` would be:

`['config', 'someProperty', 0, 'otherProperty']`

We use the path to:

1. Find the exact location of the error in the JSON document using `jsonc.findNoteAtLocation(root, nodePath)`
2. Create a diagnostic that highlights the specific part of the code where the error occurred
3. Help with deduplication of errors by using it in the `diagnosticDeduplicationKey` function

Going to resolve this comment for now but feel free to reopen with any more questions!

src/lib/projectConfigValidation.ts
```diff
@@ -0,0 +1,439 @@
...
+  } catch (error) {
+    return [
+      new Diagnostic(
+        new Range(0, 0, 0, 0),
+        getInvalidJsonError(),
+        DiagnosticSeverity.Error
+      ),
+    ];
+  }
+  const root = jsonc.parseTree(document.getText());
```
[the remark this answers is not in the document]
kemmerle: The `jsonc.parseTree` function returns a Node, not an object. AJV doesn't take Node as an input, it takes an object. Also, `jsonc` parser is very forgiving and may paper over invalid JSON. 

Going to resolve for now, feel free to reopen.

src/lib/projectConfigValidation.ts
```diff
@@ -0,0 +1,439 @@
...
+// A schema cache maps a platform version to a set of schemas and validators indexed by the internal type.
+// Validators are compiled on demand and then cached.
+type SchemaCache = Record<
+  string, // platform version
+  SchemaCacheEntry
+>;
+
+let diagnosticsCollection: DiagnosticCollection;
+let ajv: Ajv;
+let schemaCache: SchemaCache = {};
```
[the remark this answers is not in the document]
kemmerle: I got rid of the globals while keeping things functional.

src/lib/projectConfigValidation.ts
```diff
@@ -0,0 +1,439 @@
...
+async function validateDocumentDiagnostics(
+  document: TextDocument
+): Promise<Diagnostic[]> {
+  if (!document.fileName.endsWith('-hsmeta.json')) return [];
+  const projectRoot = await findProjectRoot(document.uri);
+  if (!projectRoot) return [];
+  const projectVersion = await getProjectVersion(projectRoot);
+  if (!projectVersion) return [];
+  // Any file that fails the above checks is one that we don't handle
+  let metaFile: any; // The type is whatever JSON the user provided us with
```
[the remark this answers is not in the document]
kemmerle: I will type it as `Partial<Components>`

src/lib/projectConfigValidation.ts
kemmerle: While the editor is open or until someone reloads the window. We tell them this in the error message if we fail to fetch schemas.

It would be an unpleasant editor experience to have the schema change behind your back. Imagine you didn't have an error but now you do mid-typing? Or you leave the buffer and it's in a good state, come back and there's an error because the schemas silently updated? And you did nothing to change anything? I think it would be better if schema updates happened at some obvious point in the lifecycle of the editor, like on reload or restart.

src/lib/projectConfigValidation.ts
kemmerle: Because later we will attach a validator to it: `type SchemaAndValidator = { schema: AnySchema; validator?: ValidateFunction };` on demand. We want to cache that validator. We could keep them in separate records, but then they could fall out of sync with one another if there's a bug. This way, there's one validator per schema.

src/lib/projectConfigValidation.ts
kemmerle: If we don't cache the error we will hit the API constantly every time someone types (less frequently if we debounce or only do this on open/save). The error message tells users how long it will stay cached.

```ts
const message = `${getFailedToFetchSchemasError()}. Reload your window to try again`;
```

package.json
```diff
@@ -587,15 +587,19 @@
         ]
     },
     "dependencies": {
-        "@hubspot/local-dev-lib": "^1.13.1",
+        "@hubspot/local-dev-lib": "^3.1.0",
```
camden11, line 590: I know I'm late to the party here so this may have already been discussed, but there are a lot of breaking changes between LDL v1 and v3 (for example, the return type of anything using `http` is going to be different) and from what I can tell that isn't accounted for in this PR.

It looks like you could maybe get away with not upgrading though?

TanyaScales (approved):
LGTM from the extension standpoint.

package.json
TanyaScales: Agree on this comment -- upgrading here would cause problems elsewhere and would avoid upgrading if possible. However, I do see the corresponding PR with the LDD updates - so will start reading through over there.

kemmerle: Merged in the PR to bump LDL.

PR #281, opened 2025-05-07
Make updates to imports after bumping local-dev-lib version

kemmerle:
## Description and Context
This PR needs to be merged into the PR to add JSON schema validation to the extension: #280 

In the schema validation PR, we need to bump the `@hubspot/local-dev-lib` package version to `^3.1.0` to be compatible with the new dependency `@hubspot/project-parsing-lib` library which also contains LDL as a dependency. 

There are significant differences between the two versions of LDL, and so I've made updates to the code where needed. 

**IMPORTANT NOTE PLEASE READ:**

Some imports from `@hubspot/local-dev-lib` **must** keep CommonJS syntax. This is because:

1. In `@hubspot/local-dev-lib`, the `exports` field in the `package.json` file exports the `lib` folder like this:

```
  "./*": "./lib/*.js"
```
2. Only ES modules can read the `exports` field.

3. VSCode extensions **don't support ES Modules**. [I'm not joking, this is a very active issue](https://github.com/microsoft/vscode/issues/130367). 

4. Ergo, imports from the `/lib` folder must keep CommonJS syntax. 

5. We could mess with the `tsconfig.json` file to try to change `compilerOptions.module` to `nodenext` or `node16` and override any issues, but Microsoft specifically warns against any hacks. They're brittle and prone to breakage. 

## TODO

- [ ] Address feedback
- [ ] Get more folks to test it out

## Who to Notify

@TanyaScales @brandenrodgers @camden11 @joe-yeager

### DIFF
--- a/src/lib/commands/config.ts
+++ b/src/lib/commands/config.ts
@@ -6,14 +6,13 @@ import { Portal } from '../types';
 import { portalNameInvalid } from '../validation';
 import { trackEvent } from '../tracking';
 import { showAutoDismissedStatusBarMessage } from '../messaging';
-
-const { getConfig } = require('@hubspot/local-dev-lib/config');
-const {
+import {
+  getConfig,
   deleteAccount,
   deleteConfigFile,
   renameAccount,
   updateDefaultAccount,
-} = require('@hubspot/local-dev-lib/config');
+} from '@hubspot/local-dev-lib/config';
 
 const showRenameAccountPrompt = (accountToRename: Portal) => {
   window
@@ -23,7 +22,12 @@ const showRenameAccountPrompt = (accountToRename: Portal) => {
     .then(async (newName: string | undefined) => {
       if (newName) {
         const oldName = accountToRename.name || accountToRename.portalId;
-        const invalidReason = portalNameInvalid(newName, getConfig());
+        const config = getConfig();
+        let invalidReason = '';
+        if (config) {
+          // @ts-expect-error TODO: Fix this when updating local-dev-lib
+          invalidReason = portalNameInvalid(newName, config);
+        }
 
         if (!invalidReason) {
           renameAccount(oldName, newName);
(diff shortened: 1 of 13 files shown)


src/lib/commands/config.ts
```diff
@@ -23,7 +22,12 @@ const showRenameAccountPrompt = (accountToRename: Portal) => {
     .then(async (newName: string | undefined) => {
       if (newName) {
         const oldName = accountToRename.name || accountToRename.portalId;
-        const invalidReason = portalNameInvalid(newName, getConfig());
+        const config = getConfig();
+        let invalidReason = '';
+        if (config) {
+          // @ts-expect-error TODO: Fix this when updating local-dev-lib
```
kemmerle, line 28: The deprecated config and new config types are simply incompatible with one another. We can remove this error, once we account for the new global config in the extension.

src/lib/providers/remoteFileProvider.ts
```diff
@@ -9,11 +10,12 @@ export const RemoteFileProvider = new (class
   onDidChange = this.onDidChangeEmitter.event;
 
   async provideTextDocumentContent(uri: Uri): Promise<string | undefined> {
+    showMissingAccountError();
     const filepath = uri.toString().split(':')[1];
     // filepath must be de-encoded since it gets reencoded by download in cli-lib
     const decodedFilePath = decodeURIComponent(filepath);
     try {
-      const file = await download(getAccountId(), decodedFilePath);
+      const { data: file } = await download(getAccountId()!, decodedFilePath);
```
kemmerle, line 18: This accounts for changes to the `http` module in LDL versions 3 and above.

src/lib/commands/config.ts
```diff
@@ -73,14 +77,18 @@ export const registerCommands = (context: ExtensionContext) => {
       COMMANDS.CONFIG.SELECT_DEFAULT_ACCOUNT,
       async () => {
         const config = getConfig();
+        // @ts-expect-error TODO: Fix this when updating local-dev-lib
         if (config && config.portals) {
```
brandenrodgers, line 81: Are there any utils we could leverage from LDL to get around some of these issues? Would `getConfigAccounts` help here?

PR #282, opened 2025-05-07
Bump LDL version & update imports

kemmerle:
## Description and Context
This PR needs to be merged before the PR to add JSON schema validation to the extension: #280 

In the schema validation PR, we need to bump the `@hubspot/local-dev-lib` package version to `^3.1.0` to be compatible with the new dependency `@hubspot/project-parsing-lib` library which also contains LDL as a dependency. 

There are significant differences between the two versions of LDL, and so I've made updates to the code where needed. 

**IMPORTANT NOTE PLEASE READ:**

Some imports from `@hubspot/local-dev-lib` **must** keep CommonJS syntax. This is because:

1. In `@hubspot/local-dev-lib`, the `exports` field in the `package.json` file exports the `lib` folder like this:

```
  "./*": "./lib/*.js"
```
2. Only ES modules can read the `exports` field.

3. VSCode extensions **don't support ES Modules**. [I'm not joking, this is a very active issue](https://github.com/microsoft/vscode/issues/130367). 

4. Ergo, imports from the `/lib` folder must keep CommonJS syntax. 

5. We could mess with the `tsconfig.json` file to try to change `compilerOptions.module` to `nodenext` or `node16` and override any issues, but Microsoft specifically warns against any hacks. They're brittle and prone to breakage. 

## TODO

- [x] Address feedback
- [x] Get more folks to test it out

## Who to Notify

@TanyaScales @brandenrodgers @camden11 @joe-yeager

### DIFF
--- a/src/lib/commands/config.ts
+++ b/src/lib/commands/config.ts
@@ -84,8 +84,7 @@ export const registerCommands = (context: ExtensionContext) => {
     commands.registerCommand(
       COMMANDS.CONFIG.SELECT_DEFAULT_ACCOUNT,
       async () => {
-        const defaultAccount: string | number | null | undefined =
-          getConfigDefaultAccount();
+        const defaultAccount = getConfigDefaultAccount();
         const portals: CLIAccount_DEPRECATED[] = getConfigAccounts() || [];
 
         if (portals && portals.length !== 0) {
--- a/src/lib/commands/config.ts
+++ b/src/lib/commands/config.ts
@@ -11,6 +11,8 @@ import {
   deleteConfigFile,
   renameAccount,
   updateDefaultAccount,
+  getConfigDefaultAccount,
+  getConfigAccounts,
 } from '@hubspot/local-dev-lib/config';
 import { CLIConfig } from '@hubspot/local-dev-lib/types/Config';
 import { CLIAccount_DEPRECATED } from '@hubspot/local-dev-lib/types/Accounts';
@@ -82,17 +84,18 @@ export const registerCommands = (context: ExtensionContext) => {
     commands.registerCommand(
       COMMANDS.CONFIG.SELECT_DEFAULT_ACCOUNT,
       async () => {
-        const config: CLIConfig | null = getConfig();
+        const defaultAccount: string | number | null | undefined =
+          getConfigDefaultAccount();
+        const portals: CLIAccount_DEPRECATED[] = getConfigAccounts() || [];
 
-        if (config && 'portals' in config && config.portals) {
+        if (portals && portals.length !== 0) {
           window
             .showQuickPick(
-              config.portals.map((p: CLIAccount_DEPRECATED) => {
+              portals.map((p: CLIAccount_DEPRECATED) => {
                 return {
                   label: getDisplayedHubspotPortalInfo(p),
                   description:
-                    config.defaultPortal === p.portalId ||
-                    config.defaultPortal === p.name
+                    defaultAccount === p.portalId || defaultAccount === p.name
                       ? '(default)'
                       : '',
                   portal: p,
@@ -138,7 +141,7 @@ export const registerCommands = (context: ExtensionContext) => {
     commands.registerCommand(
       COMMANDS.CONFIG.DELETE_ACCOUNT,
       async (accountToDelete) => {
-        const config = getConfig();
+        const portals: CLIAccount_DEPRECATED[] = getConfigAccounts() || [];
         const accountIdentifier =
           accountToDelete.name || accountToDelete.portalId;
 
@@ -150,11 +153,7 @@ export const registerCommands = (context: ExtensionContext) => {
           )
           .then(async (answer) => {
             if (answer === 'Yes') {
-              if (
-                config &&
-                'portals' in config &&
-                config.portals.length === 1
-              ) {
+              if (portals && portals.length === 1) {
                 deleteConfigFile();
                 showAutoDismissedStatusBarMessage(
                   `Successfully deleted account ${accountIdentifier}. The config file has been deleted because there are no more authenticated accounts.`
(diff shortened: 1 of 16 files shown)


src/lib/commands/config.ts
```diff
@@ -23,7 +22,12 @@ const showRenameAccountPrompt = (accountToRename: Portal) => {
     .then(async (newName: string | undefined) => {
       if (newName) {
         const oldName = accountToRename.name || accountToRename.portalId;
-        const invalidReason = portalNameInvalid(newName, getConfig());
+        const config = getConfig();
+        let invalidReason = '';
+        if (config) {
+          // @ts-expect-error TODO: Fix this when updating local-dev-lib
```
kemmerle, line 28: The deprecated config and new config types are simply incompatible with one another. We can remove this error, once we account for the new global config in the extension.

src/lib/commands/remoteFs.ts
```diff
@@ -26,6 +30,7 @@ export const registerCommands = (context: ExtensionContext) => {
     commands.registerCommand(
       COMMANDS.REMOTE_FS.FETCH,
       async (clickedFileLink) => {
+        showMissingAccountError();
```
kemmerle, line 33: If no account is specified, we throw an error to the user and exit the command. This guarantees that `getAccountId()` is not undefined.

src/lib/providers/remoteFileProvider.ts
```diff
@@ -9,11 +10,12 @@ export const RemoteFileProvider = new (class
   onDidChange = this.onDidChangeEmitter.event;
 
   async provideTextDocumentContent(uri: Uri): Promise<string | undefined> {
+    showMissingAccountError();
     const filepath = uri.toString().split(':')[1];
     // filepath must be de-encoded since it gets reencoded by download in cli-lib
     const decodedFilePath = decodeURIComponent(filepath);
     try {
-      const file = await download(getAccountId(), decodedFilePath);
+      const { data: file } = await download(getAccountId()!, decodedFilePath);
```
kemmerle, line 18: This accounts for changes to the http module in LDL versions 3 and above.

src/lib/commands/config.ts
```diff
@@ -73,14 +77,18 @@ export const registerCommands = (context: ExtensionContext) => {
       COMMANDS.CONFIG.SELECT_DEFAULT_ACCOUNT,
       async () => {
         const config = getConfig();
+        // @ts-expect-error TODO: Fix this when updating local-dev-lib
         if (config && config.portals) {
```
brandenrodgers, line 81: Just porting this from the old PR:

Are there any utils we could leverage from LDL to get around some of these issues? Would getConfigAccounts help here?

src/lib/commands/config.ts
brandenrodgers: Would it help to import the config types from LDL? iirc it has some types that account for both config shapes

src/lib/uri.ts
```diff
@@ -16,19 +16,28 @@ const {
...
 } = require('@hubspot/local-dev-lib/personalAccessKey');
-const {
+import {
   createEmptyConfigFile,
   setConfigPath,
-} = require('@hubspot/local-dev-lib/config');
+} from '@hubspot/local-dev-lib/config';
+
+// TODO: Remove these once we've updated local-dev-lib
+const ENVIRONMENTS = {
```
brandenrodgers, line 25: Can you clarify the todo here? LDL should have an ENVIRONMENTS export.

src/lib/helpers.ts
```diff
@@ -99,3 +101,13 @@ export const buildStatusBarItem = (text: string) => {
   statusBarItem.text = text;
   return statusBarItem;
 };
+
+export function showMissingAccountError() {
```
brandenrodgers, line 105: I like this idea, but I'm not sure what to do to get it to show. I removed my hubspot.config.yml file but I'm not seeing this error surfacing anywhere. Do you know a way for me to test it?

brandenrodgers:
We may need to take a look at how this is handling error messages since those have changed in the newer LDL version too. Here's an example of one: 
image

**You can reproduce this by:**
1. Removing your hubspot.config.yml file
2. Opening the HubSpot left pane in the extension
3. Starting a watcher by clicking the menu icon next to "Remote file system" (then choose any file or folder to watch)
4. Edit that file/folder and save

I tested on the master branch and this is the error that shows:
image

src/lib/commands/config.ts
```diff
@@ -142,8 +150,11 @@ export const registerCommands = (context: ExtensionContext) => {
           )
           .then(async (answer) => {
             if (answer === 'Yes') {
-              // @ts-expect-error TODO: Fix this when updating local-dev-lib
-              if (config && config.portals.length === 1) {
+              if (
+                config &&
```
brandenrodgers, line 154: Could this be cleaned up with `getConfigAccounts()`?

src/lib/commands/config.ts
```diff
@@ -76,19 +82,16 @@ export const registerCommands = (context: ExtensionContext) => {
     commands.registerCommand(
       COMMANDS.CONFIG.SELECT_DEFAULT_ACCOUNT,
       async () => {
-        const config = getConfig();
-        // @ts-expect-error TODO: Fix this when updating local-dev-lib
-        if (config && config.portals) {
+        const config: CLIConfig | null = getConfig();
+
+        if (config && 'portals' in config && config.portals) {
```
brandenrodgers, line 87: Similar question here for `getConfigAccounts()`

src/lib/commands/config.ts
```diff
@@ -72,16 +84,18 @@ export const registerCommands = (context: ExtensionContext) => {
     commands.registerCommand(
       COMMANDS.CONFIG.SELECT_DEFAULT_ACCOUNT,
       async () => {
-        const config = getConfig();
-        if (config && config.portals) {
+        const defaultAccount: string | number | null | undefined =
+          getConfigDefaultAccount();
```
brandenrodgers, line 88: `getConfigDefaultAccount` has its own return type, so it's redundant to declare the types for `defaultAccount` here.

brandenrodgers:
This is looking good to me! Just had one comment about an unnecessary type. I'll let @TanyaScales give the final ✅ on this since she's more familiar with it.

TanyaScales (approved):

PR #287, opened 2025-06-11
Update VS code to be compatible with the global config

brandenrodgers:
This leverages all the utils from LDL that help with global config compatibility. The tldr is that we shouldn't directly reference `portalId` or `defaultAccount` anymore.

I also made a fix to the account delete and rename actions. They work correctly, but they weren't refreshing the accounts section after performing the updates which makes it look like it isn't working.

### DIFF
--- a/src/lib/commands/config.ts
+++ b/src/lib/commands/config.ts
@@ -1,10 +1,4 @@
 import { commands, window, ExtensionContext } from 'vscode';
-import { updateStatusBarItems } from '../statusBar';
-import { COMMANDS, TRACKED_EVENTS } from '../constants';
-import { getDisplayedHubspotPortalInfo } from '../helpers';
-import { portalNameInvalid } from '../validation';
-import { trackEvent } from '../tracking';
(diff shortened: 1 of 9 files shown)


src/lib/commands/config.ts
```diff
@@ -24,7 +28,8 @@ const showRenameAccountPrompt = (accountToRename: CLIAccount_DEPRECATED) => {
     })
     .then(async (newName: string | undefined) => {
       if (newName) {
-        const oldName = accountToRename.name || accountToRename.portalId;
+        const oldName =
+          accountToRename.name || getAccountIdentifier(accountToRename);
```
brandenrodgers, line 32: getAccountIdentifier() will look for `account.portalId || account.accountId`

camden11 (approved):
Assuming you'll want a review from someone from CMS dev tooling, but from the config side of things everything here LGTM

TanyaScales:
Tested and that is smooth! Thank you - Looks great and worked to migrate and merge my config files

TanyaScales (approved):

PR #295, opened 2025-10-17
Reorganizing + fixing hubl auto detection

brandenrodgers:
Some mostly structural updates to the project to make it easier to build on. This makes some fixes to the hubl auto detection functionality that seems to have not been working. The major changes:
- hubl auto detection should work now (I could have just been testing this incorrectly, but it wasn't working for me originally)
- Reorganizing the folder structure to be a bit more intuitive
- Grouping common utils into files in lib/
- Moving features out of the lib folder
- Creating a dedicated "Events" folder
- Turning certain registered commands into standard js functions in areas where they don't need to be registered
- Cleaning up some dead code
- Removing some unnecessary deps
- Organizing imports for all files

### DIFF
--- a/package.json
+++ b/package.json
@@ -527,12 +527,12 @@
         "viewsWelcome": [
             {
                 "view": "hubspot.viewsWelcome.accountAuth",
-                "contents": "No authenticated accounts were found.\nIn order to use HubSpot VSCode features, you need to authenticate your account(s). \n[Authenticate HubSpot Account](command:hubspot.auth.onClickAuthorize)\nTo learn more about authentication using personal access keys, [read our docs](https://developers.hubspot.com/docs/cms/personal-access-key).",
+                "contents": "No authenticated accounts were found.\nIn order to use HubSpot VSCode features, you need to authenticate your account(s). \n[Authenticate HubSpot Account](command:hubspot.account.onClickAuthorize)\nTo learn more about authentication using personal access keys, [read our docs](https://developers.hubspot.com/docs/cms/personal-access-key).",
                 "when": "!hubspot.configPath"
             },
             {
                 "view": "hubspot.viewsWelcome.accountAuth",
-                "contents": "To work with other HubSpot accounts within VSCode, authenticate each additional account by clicking the button below. \n[Authenticate Additional HubSpot Account](command:hubspot.auth.onClickAuthorize)\nTo learn more about authentication using personal access keys, [read our docs](https://developers.hubspot.com/docs/cms/personal-access-key).",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -296,21 +292,9 @@
...
             },
-            {
-                "command": "hubspot.remoteFs.refresh",
-                "title": "Cached Refresh"
-            },
             {
                 "command": "hubspot.remoteFs.hardRefresh",
-                "title": "Refresh"
-            },
-            {
```
brandenrodgers: These commands can be deleted from here. This project uses a pattern where we have "events" and "commands".
- **commands** - user-facing commands that can be executed via some form of ui interaction
- **events** - internal-only "commands" that cannot be directly triggered by a user through the ui

Anything that is purely an event, does not need to be configured here in the package.json file. I cleaned up everything in here that was defined as a command, has a "when: false" clause, and is not referenced anywhere else in the package.json file (i.e. menu options or something else).

src/lib/commands/globalState.ts
```diff
@@ -1,20 +0,0 @@
-import { commands, ExtensionContext } from 'vscode';
-import * as dayjs from 'dayjs';
-import { COMMANDS } from '../constants';
-
-export const registerCommands = (context: ExtensionContext) => {
```
brandenrodgers: Per the patterns of this project, this is actually an "event" instead of a command. But it also isn't leveraged anywhere in the package.json file, so we can remove this completely. It's just a thin wrapper around the `context.globalState.update()` function. I added two utils to the helpers.ts file, updated the `panels/feedback.ts` file, and then updated all the code that previously called `executeCommand(COMMANDS.GLOBAL_STATE.UPDATE_DELAY)` to directly call the global state update function instead.

src/events/index.ts
```diff
@@ -0,0 +1,11 @@
+import { ExtensionContext } from 'vscode';
```
brandenrodgers: I'm still ramping up here, but from what I can tell so far this "Events" pattern is not a good one. In most (all?) cases there isn't a need to leverage VSCode's event bus to execute logic like this. AFAIK the primary reason to register events is to expose them as user-facing functionality in the `contributes` section of the package.json. These events are adding unnecessary complexity to the logic flow by leveraging the event bus when they don't need to be.

I undid some of it (all the terminal logic that checks the installation status for npm and the cli, and also the global state events), but not all of it because some events are pretty deeply rooted.

src/lib/constants.ts
```diff
@@ -1,13 +1,3 @@
-export const TEMPLATE_ERRORS_TYPES = {
```
brandenrodgers: I removed some constants from here that are only used in a single place

camden11 (commented):
Once again, not really a VSCode expert but code here LGTM. Only issue I can find is the auth button not working

brandenrodgers:
@camden11 good catch. I fixed the key in [de9debe](https://github.com/HubSpot/hubspot-cms-vscode/pull/295/commits/de9debe22938b65f011e9eb5f8185d5e58986bfe)

PR #306, opened 2025-11-14
Format feedback form

brandenrodgers:
### Before
image

### After
image


chiragchadha1:
Is it possible to make invert the colors when the IDE has a dark theme? It'd be nice if the form background matched theme background and the text was white.

brandenrodgers:
@chiragchadha1 good idea! Here's an update where it now follows the IDE's theme

image
image
image

PR #338, opened 2026-04-22
Fix imports from project-parsing-lib

brandenrodgers:
## Description and Context

We bumped the PPL dep from 0.10.3 to 0.15.0, which picked up some changes to the exports.

I also fixed an infinite loop in the config migration util that would perpetually think that it needs to migrate a local config to global.

The rest are simple changes to get the build to pass.


src/lib/config.ts
```diff
@@ -90,6 +87,7 @@ export const loadHubSpotConfigFile = (rootPath: string) => {
               GlobalConfigWithPropertiesMerged,
               deprecatedConfig!
             );
+            archiveConfigAtPath(deprecatedConfigPath);
```
brandenrodgers, line 90: This was causing the loop during the config merge flow. We need to archive the local config so the extension ignores it.

tsconfig.json
```diff
@@ -7,7 +7,8 @@
     "sourceMap": true,
     "strict": true,
     "rootDir": "src",
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "esModuleInterop": true
```
brandenrodgers, line 11: This is for a webpack build error related to the default import for dayjs
```bash
[tsl] ERROR in /Users/brodgers/src/public/hubspot-cms-vscode/src/lib/helpers.ts(2,8)
      TS1259: Module '"/Users/brodgers/src/public/hubspot-cms-vscode/node_modules/dayjs/index"' can only be default-imported using the 'esModuleInterop' flag
```

chiragchadha1 (approved):
LGTM

PR #349, opened 2026-06-11
fix: Extension fails to load in windows due to path issue

joe-yeager:
## Description and Context

Users on Windows were seeing **"File URL path must be absolute"** when the extension loaded. The root cause is `import.meta.url` in LDL's `handleFieldsJS.js` being evaluated at module scope — webpack inlines it as the build machine's absolute file URL, which `fileURLToPath()` rejects on Windows because the Unix-style path isn't valid there.

### Changes

- **Bumped `@hubspot/local-dev-lib` 5.0.0 → 5.8.2** — includes the fix that moves `import.meta.url` inside the function body so it is no longer evaluated at module load time
- **Added `@aws-sdk/client-s3` as a webpack external** — prevents the AWS SDK from being bundled, resolving a separate load failure on Windows

## Who to Notify


webpack.config.js
```diff
@@ -27,6 +27,7 @@ const config = {
   externals: [
     {
       vscode: 'commonjs vscode',
+      '@aws-sdk/client-s3': 'commonjs @aws-sdk/client-s3',
```
joe-yeager, line 38: This is an optional dependency that was added as a part of adding unzipper to LDL.  We don't need the aws capabilities, so adding this as external so it doesn't get included in the bundle because it was causing errors.

package.json
```diff
@@ -18,7 +18,8 @@
         "webpack": "webpack --mode development",
         "webpack-dev": "webpack --mode development --watch",
         "test-compile": "tsc -p ./",
-        "release": "yarn tsx ./scripts/release.ts release"
+        "release": "yarn tsx ./scripts/release.ts release",
+        "prerelease:package": "npx --yes @vscode/vsce package --pre-release -o releases/"
```
joe-yeager, line 22: Just adding this as a utility to make testing easier.  We can bundle the .vsix file and install it locally using that file.

chiragchadha1 (approved):
Thanks for looking into this!

PR #361, opened 2026-06-17
test: Add acceptance tests via VS Code Extension Development Host

joe-yeager:
## Description and Context

Adds an `acceptance-test/` suite that runs inside the real VS Code Extension Development Host using `@vscode/test-cli` + `@vscode/test-electron`. Tests drive VS Code's actual completion pipeline via `vscode.executeCompletionItemProvider` rather than calling provider classes directly — this means provider registration, trigger character wiring, language scoping, and filesystem scanning are all exercised end-to-end.

**What's tested:**
- Extension activates and all 20 `package.json` commands are registered in the VS Code command registry
- `FileCompletionProvider` trigger characters (`'` and `"`) are wired up for `html-hubl`
- All four HubL keyword patterns (`include`, `extends`, `import`, `path=`) trigger file completions
- Non-matching lines produce no `File`-kind completions (guard logic verified in the real pipeline)
- Provider is scoped to `html-hubl` only — a `css-hubl` document returns zero file completions
- Completion items reference actual workspace fixture files, use `./`-prefixed relative paths, and correctly append/omit the closing quote based on line suffix

**Key implementation note:** VS Code's document model is decoupled from the filesystem — writing to an open file's path does not update the in-memory model. All tests use `WorkspaceEdit.replace` to patch document content directly, avoiding stale-content failures between tests.

Run with: `yarn acceptance-test`

## Screenshots

N/A — no UI changes.

## TODO

- Consider wiring `yarn acceptance-test` into CI (requires `Xvfb` on Linux runners)


acceptance-test/fixtures/workspace/module.html
```diff
@@ -0,0 +1 @@
+<!-- HubSpot module fixture for integration tests -->
```
joe-yeager, line 1: These files are empty, but are added here so that the acceptance tests can test the inline completion for filenames

acceptance-test/suite/extension.test.ts
```diff
@@ -0,0 +1,64 @@
+import * as assert from 'assert';
+import * as vscode from 'vscode';
+
+const EXTENSION_ID = 'HubSpot.hubl';
+
+suite('Extension', () => {
```
joe-yeager, line 6: This test suite is verifying that we are correcting activating the extension and registering our commands correctly with VSCode.

chiragchadha1 (approved):
Thanks for adding!

PR #363, opened 2026-06-17
test: Add acceptance test suites for providers, snippets, linting, and remoteFs scheme

joe-yeager:
## Description and Context

Adds five new acceptance test suites to `acceptance-test/suite/` covering parts of the extension that are difficult to unit-test without the real VS Code Extension Development Host:

- **`accountsTreeDataProvider.test.ts`** — verifies `getChildren` returns the fixture account (id 99999), `getTreeItem` shape/icon/label, and empty-accounts edge case via a temp config
- **`docsTreeDataProvider.test.ts`** — verifies 6 children are returned, URL items get `url-link-tree-item`/`link-external`, and the feedback item gets `command-tree-item` with the correct command ID
- **`hublSnippets.test.ts`** — verifies `~`-prefixed tag snippets and `content`-prefixed variable snippets appear in `html-hubl` files, and that HubL snippets are absent in plain `.html` files
- **`remoteFileContentProvider.test.ts`** — verifies the `hubl://` scheme is registered (opening a `hubl://` URI doesn't throw "no text content provider")
- **`linting.test.ts`** — verifies no HubL diagnostics on `.ts` files, and that diagnostics are cleared when switching away from `html-hubl`

Supporting changes: removed `rootDir` from `acceptance-test/tsconfig.json` (required to allow cross-directory imports from `src/`), added `--user-data-dir /tmp/vsc-ext-test` to `.vscode-test.mjs` to avoid the 103-char Unix socket path limit when running from deep worktree paths.

## Screenshots

N/A — all changes are test infrastructure.


chiragchadha1 (approved):
LGTM!

